PR #14, opened 2019-03-20
Fix music player bugs

iisa:
- Fix Webamp link hover state in music player (webdev-2058)
- Update Music Player wrapper to show full track length as option (webdev-2055)
- Theatre viewer styling with tabs (webdev-2052)
- Music Player Channel Selector: programmatically display external channel labels correctly (webdev-2061)
- Music Player: Removing track numbers from track list when showing external channels (webdev-2059)


packages/ia-components/sandbox/pagination/pagination-with-flexbox.jsx
```diff
@@ -239,7 +239,12 @@ class Paginator extends Component {
     if (numberOfPages === pageSelected) return null;
 
     return (
-      <button type="button" className="pagination-arrow right" onClick={this.goToNextPage}>
+      <button
+        type="button"
+        className="pagination-arrow right"
+        onClick={this.goToNextPage}
+        data-event-click-tracking="Paginator|Arrow-Right"
```
mekarpeles: Does this need an aria label for accessibility?

packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify-main.jsx
```diff
@@ -212,13 +221,14 @@ class AudioPlayerWithYoutubeSpotify extends Component {
...
     } = albumData;
     let audioPlayerChannelLabel;
-    if (channelToPlay === 'archive') {
-      audioPlayerChannelLabel = `Archive${playSamples ? ' Samples' : ''}`;
+    const isArchiveChannel = channelToPlay === 'archive';
+    if (isArchiveChannel) {
+      audioPlayerChannelLabel = `${playSamples ? ' Samples' : 'Internet Archive'}`;
     } else {
-      audioPlayerChannelLabel = channelToPlay;
+      audioPlayerChannelLabel = externalSourcesDisplayValues[channelToPlay] || '';
```
mekarpeles: Where is `channelToPlay` defined -- is it possible to be undefined?

packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify-main.jsx
```diff
@@ -261,6 +271,8 @@ class AudioPlayerWithYoutubeSpotify extends Component {
             onSelected={this.selectThisTrack}
             selectedTrack={trackSelected}
             albumName={title[0]}
+            displayTrackNumbers={isArchiveChannel}
+            creator={creator[0]}
```
mekarpeles: can't creator be `''` above? Does this need to be something like `creator.length ? creator[0] : ''`?

packages/ia-components/variables.less
```diff
@@ -5,5 +5,6 @@
 @audio-player-desktop-min-width: 768px;
 @audio-player-desktop-break: 950px;
 
+@width-500: 500px;
 // colors
 @f-white: #fff;
```
mekarpeles: newline end of file? Out of scope, but possible improvement

packages/ia-components/sandbox/pagination/pagination-with-flexbox.jsx
iisa: @koderjoker would like your input on this block of code

koderjoker: I agree that the next, previous and page number buttons need aria labels since otherwise they would just read out as "button" on screen readers. We could give our next and previous buttons aria-labels "Go to next page" and "Go to previous page" respectively, and give the page buttons the aria-label={`Go to page number ${isPage ? 'selected' : ''}`}

koderjoker: Additionally, (if we want to also tell a user what page they are on) we could change the above to aria-label={isPage ? `Current page, page ${thisPage}` : `Go to page ${thisPage}`}

koderjoker: Though I would say that it is mainly the page buttons that require an aria-label, since the next and previous button hold a somewhat reasonable description.

packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify-main.jsx
iisa: good catch, it's has a default value, won't be undefined

packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify-main.jsx
iisa: the incoming data has been normalized so values come in as arrays.  Thanks for the catch

packages/ia-components/sandbox/pagination/pagination-with-flexbox.jsx
bfalling: @iisa: Separate comment re: click tracking. I know that this Paginator component is generic, and so the click tracking sticks with a generic Category (the first value), but we might want to think about this a little bit. If we have multiple instances of the Paginator, then it's going to be difficult to know which Paginator people clicked on. For the current use case, we want to know how many times the *Music Links* Paginator was clicked. Not sure if there's a good way to pass in the context into the Category value and then use the Action value for the specific affordance, e.g. "Paginator-Arrow-Right". This comment applies to all the click tracking.

iisa: @bfalling good catch.

Issue #16, opened 2019-03-22
Installation steps fails for Ubuntu 18.04


shreyansh23:
On installing the project using the instructions in README.md. The steps fails at some points.
This includes:
- error due to fsevents package
- babal-core version error

Batey96:
exactly what i faced.Did  some research and found out fsevent no more supported on linux Os.
To erase the yarn instance warning,simply run the instances parralelly with this on the terminal:
**yarn run lerna bootstrap --concurrency=1*

Batey96:
exactly what i faced.Did some research and found out fsevent no more supported on linux Os.
To erase the yarn instance warning,simply run the instances parralelly with this on the terminal:
**yarn run lerna bootstrap --concurrency=1**

shreyansh23:
Yes @Batey96 `fsevents` do not support linux. The correct way of using that module is putting it in optional dependencies. Please see the PR. I shared the relevant resources in the comments.

shreyansh23:
Issue solved in #22
Closing this issue.

PR #19, opened 2019-03-22
Move fsevents to optional dependencies in package.json

shreyansh23:
Related to #16 

The package "fsevents" was in written in dependencies in package.json `ia-components`. But that package is only compatible for macOS.

Ref: https://medium.com/@satawneh/npm-install-error-fsevents-in-ohter-mac-platforms-953101fc2797

Because of this error was coming on `yarn install` on Linux operating systems.

After this PR all installation steps will work correctly for macOS as well as for Linux.


shreyansh23:
No @Batey96 
I didn't get the following that errors. Can you please pull my changes and then try the exact instruction mentioned in README.md.
I am pretty sure that you will not get any problem.

shreyansh23:
> @shreyansh23 yes let's update all the babel versions in the other repos to stay in parity. would you also check if we actually need @fsevents in `ia-components`?

Okay @iisa 
I am upgrading all the babel packages.

shreyansh23:
Got closed by mistake. Please see #22 .

mitra42:
Careful with Babel - it seems that none of us really understand it, and there are incompatibilities with what you have to include in what config files (babel.rc and webpack.config) for different versions.  I have a complete mess in the IAUX2 branch of dweb-archive and cannot find any combination of babel that will run properly with webpack and also at runtime. On Friday @iisa and I got it compiling with up to date babel, but it complained about a runtime missing - adding the runtime meant it wouldn't compile.

mitra42:
If you want to try this .... 
Check out [dweb-mirror](https://github.com/internetarchive/dweb-mirror)#iaux2 and checkout [iaux](https://github.com/internetarchive/iaux)#dwebnew. They should be in side-by-side directories (e.g. `git/dweb-mirror` and `git/iaux` ). Then try `cd dweb-mirror && webpack --mode development`.  I cannot get it to webpack except by removing some things, in particular `plugin-transform-runtime` and it seems to need one of those plugins to run in the browser (Its not clear if its `babel-plugin-transform-runtime` or `@babel/plugin-transform-runtime`). 

Note - you'll probably need to do `cd dweb-mirror && yarn install` and also maybe `cd dweb-mirror && ./install.sh` before webpack-ing. 

If you figured out how to get this to work it would be huge.

PR #22, opened 2019-03-24
Fix Linux installation problem, move fsevents to optional dependencies

shreyansh23:

Fix issue issue in Linux installation + Migrate to Babel 7 :tada: 
Related to PR #19 
Closes issue #16 

- Earlier there were problems in Linux installation. The problem was due to `fsevents` package.
- The babel version 6 was used. But the testing required version 7. So, the tests were failing! ([Ref](https://github.com/internetarchive/iaux/pull/19#issuecomment-475745972)).

**Solution**
- `fsevents` is moved to `optionalDependencies` which is where it should be present. 
- Migrate `Babel` from version 6 to 7

**Implementation**
Implemenation was done following the steps suggested by https://babeljs.io/docs/en/configuration

n/a

Can be tested by pulling the branch. Then follow the steps suggested in the [README.md](https://github.com/internetarchive/iaux)
Steps:
```yarn install
yarn run lerna bootstrap
yarn run lerna link
cd packages/ia-components
yarn run storybook
```

Screenshot from 2019-03-24 18-09-46

Screenshot from 2019-03-24 19-28-05

Screenshot from 2019-03-24 18-11-13

### DIFF
--- a/packages/ia-components/package.json
+++ b/packages/ia-components/package.json
@@ -23,7 +23,6 @@
         "assert": "^1.4.1",
         "babel-loader": "^8.0.5",
         "css-loader": "^2.1.0",
-        "fsevents": "^1.2.4",
         "jest": "^23.6.0",
         "less": "^3.8.1",
         "less-loader": "^4.1.0",
@@ -39,6 +38,9 @@
     },
     "peerDependencies": {
         "@internetarchive/ia-js-client": "^0.0.0-alpha.20+5039f8d"
+    },
+    "optionalDependencies": {
+        "fsevents": "^1.2.4"
     },
     "publishConfig": {
         "access": "public"


shreyansh23:
@bfalling @iisa @mekarpeles 
Please review the PR

mitra42:
Careful with Babel - it seems that none of us really understand it, and there are incompatibilities with what you have to include in what config files (babel.rc and webpack.config) for different versions. I have a complete mess in the IAUX2 branch of dweb-archive and cannot find any combination of babel that will run properly with webpack and also at runtime. On Friday @iisa and I got it compiling with up to date babel, but it complained about a runtime missing - adding the runtime meant it wouldn't compile.

mitra42:
If you want to try this ....
Check out dweb-mirror#iaux2 and checkout iaux#dwebnew. They should be in side-by-side directories (e.g. git/dweb-mirror and git/iaux ). Then try cd dweb-mirror && webpack --mode development. I cannot get it to webpack except by removing some things, in particular plugin-transform-runtime and it seems to need one of those plugins to run in the browser (Its not clear if its babel-plugin-transform-runtime or @babel/plugin-transform-runtime).

Note - you'll probably need to do cd dweb-mirror && yarn install and also maybe cd dweb-mirror && ./install.sh before webpack-ing.

If you figured out how to get this to work it would be huge.

packages/ia-design-system/webpack.config.js
```diff
@@ -39,7 +39,7 @@ module.exports = {
           loader: 'babel-loader',
           options: {
             babelrc: false,
-            presets: ['es2015'],
+            presets: ['@babel/preset-env'],
```
iisa: 👍

packages/ia-components/package.json
```diff
@@ -21,9 +21,10 @@
         "@storybook/addon-knobs": "^4.1.4",
         "@storybook/react": "^4.1.4",
         "assert": "^1.4.1",
+        "babel-core": "^7.0.0-bridge.0",
```
iisa: Question: why are we using `babel-core` instead of `@babel/core` v. 7?

iisa: we already use ` "@babel/core": "^7.2.2",` at the first line, we should not need the two unless it is for a certain reason.

shreyansh23: @iisa 
Your concern is valid. I will need to check with docs and `babel-upgrade` package which I used for upgrading the version. I will do some research and reply.

shreyansh23: @iisa I think this happened because `ia-components/package.json` was already using Babel 7 and `babel-upgrade` package was I used was designed to migrate from Babel 6 to Babel 7. So changes in this `package.json` were not required. Thanks for reviewing. 
I will update the PR.

shreyansh23:
> If you want to try this ....
> Check out dweb-mirror#iaux2 and checkout iaux#dwebnew. They should be in side-by-side directories (e.g. git/dweb-mirror and git/iaux ). Then try cd dweb-mirror && webpack --mode development. I cannot get it to webpack except by removing some things, in particular plugin-transform-runtime and it seems to need one of those plugins to run in the browser (Its not clear if its babel-plugin-transform-runtime or @babel/plugin-transform-runtime).
> 
> Note - you'll probably need to do cd dweb-mirror && yarn install and also maybe cd dweb-mirror && ./install.sh before webpack-ing.
> 
> If you figured out how to get this to work it would be huge.

@mitra42 
I will definitely try that.

traceypooh:
@mitra42 well, _technically speaking_ i understand babel very well.
i set us up with it ~4y+ ago; 
then moved us from a "push out" script to package.json script;
then helped move it to webpack

the problem has been more in last ~2y we haven't been as rigorous about testing updates -- IE esp. got a bit tricky some additional polyfill requirements, etc.

shreyansh23:
@traceypooh Would you also please review the PR. I think we should leave no stone unturned.

mitra42:
Glad to hear that @traceypooh, we didn't think to ask you. 

Even on stackexchange people dont seem to fully understand what versions of what packages are needed for current versions, and the problem seems to be that babel does a crappy job of defining dependencies so if you don't have just the right set it gives incomprehensible error messages.

packages/ia-components/package.json
iisa: Jest JS docs say otherwise.  Perhaps upgrade tool is behind.  Pleaes double check. https://jestjs.io/docs/en/getting-started

Screen Shot 2019-03-25 at 9 18 15 AM

packages/ia-components/yarn.lock
```diff
@@ -2077,6 +2077,11 @@ babel-core@^6.0.0, babel-core@^6.26.0:
     slash "^1.0.0"
     source-map "^0.5.7"
 
+babel-core@^7.0.0-bridge.0:
```
iisa: please note to update yarn.lock once this is resolved

packages/ia-prototype-apps/package.json
```diff
@@ -4,19 +4,20 @@
   "license": "AGPL-3.0-only",
   "private": true,
   "dependencies": {
-    "babel-preset-react": "^6.24.1",
+    "@babel/core": "^7.0.0",
```
iisa: @babel/core has been added twice, here and in devDependencies.

packages/ia-components/package.json
shreyansh23: @iisa I again cross-checked. 
It is strange but `babel-core` is **needed** in case of jestJS.
Relevant links:
1: https://github.com/facebook/jest/issues/6913#issuecomment-417123948
2: https://github.com/facebook/jest/issues/6662#issuecomment-403718906
All the the solutions say the same thing, i.e. keep both @babel/core and babel-core in case of jest as a dependency.
So I strongly think that the `babel-upgrade` is doing the right thing.
Also if you see link 1, removing `babel-core` gives errors.
So concluding on the note that `babel-upgrade` is doing the right thing in my opinion.

iisa:
@shreyansh23 hi, can you pull the FSEvents to optional dependency as its own PR?

shreyansh23:
Updating this PR.
This PR only fixed move `fsevents` to optional dependencies.
Babel Migration is handled by PR #51

iisa:
Babel upgrade has been moved to a different PR #51

PR #23, opened 2019-03-24
Correct "test:generate-output command" in package.json

### LINKED ISSUE #21: Jest tests failing on second time running storybook
**Description**
> What problem are we solving? What does the experience look like today? What are the symptoms? Please describe the component/story/add on related to the issue. 

The test running command was causing a problem on second time running `yarn run storybook`.

`   "test:generate-output": "mkdir jest-test-utils && jest --json --outputFile=jest-test-utils/jest-test-results.json || true",`

But if `jest-test-utils` folder already exist the test then the above mentioned command was unable to run.

**Evidence / Screenshot (if possible)**
Screenshot from 2019-03-24 18-03-43

**Expectation**
> What should by happening? What will it look like / how will it behave?

The command should work properly.

**Details**

> Browser type and version? Firefox latest

> Operating system? Ubuntu 18.04

**Proposal & Constraints**

> What is the proposed solution / implementation? Is there a precedent of this approach succeeding elsewhere?

Adding `rm -rf jest-test-utils` in the `test:generate-output`.

**Stakeholders**
> @ tag stakeholders of this bug

@iisa @bfalling @mekarpeles

shreyansh23:

Closes #21 

n/a

Run `yarn run storybook` as many times. The jest tests will work fine.

Screenshot from 2019-03-24 18-25-25

### DIFF
--- a/packages/ia-components/package.json
+++ b/packages/ia-components/package.json
@@ -6,7 +6,7 @@
     "scripts": {
         "storybook": "yarn run test:generate-output && yarn run storybook:start",
         "storybook:start": "start-storybook -p 9001 -c .storybook",
-        "test:generate-output": "mkdir jest-test-utils && jest --json --outputFile=jest-test-utils/jest-test-results.json || true",
+        "test:generate-output": "rm -rf jest-test-utils && mkdir jest-test-utils && jest --json --outputFile=jest-test-utils/jest-test-results.json || true",
         "test:update-snapshots": "jest --updateSnapshot",
         "test": "jest --watch --coverage"
     },


packages/ia-components/package.json
```diff
@@ -6,7 +6,7 @@
     "scripts": {
         "storybook": "yarn run test:generate-output && yarn run storybook:start",
         "storybook:start": "start-storybook -p 9001 -c .storybook",
-        "test:generate-output": "mkdir jest-test-utils && jest --json --outputFile=jest-test-utils/jest-test-results.json || true",
+        "test:generate-output": "rm -rf jest-test-utils && mkdir jest-test-utils && jest --json --outputFile=jest-test-utils/jest-test-results.json || true",
```
iisa: once we have a whole lot more tests, removing this may take a while. The error is a moot one as it is not a hard fail, the yarn command continues to move forward running storybook.  A suggestion would be to look for the `jest-test-utils` directory and if it isn't there, create it. This can be parsed out into its own script so its usage is composable like the others.

shreyansh23: Yes, @iisa I will seek a better implementation as suggested by you.

PR #46, opened 2019-03-28
Add tiles components and IAReactComponent from dweb-archive

mitra42:

Closes #

### DIFF
--- a/packages/ia-components/sandbox/IAReactComponent.js
+++ b/packages/ia-components/sandbox/IAReactComponent.js
@@ -1,0 +1,31 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+const debug = require('debug')('dweb-archive:ConfigDetailsComponent');
+
+export default class IAReactComponent extends React.Component {
+    // Both dweb-archive.IAFakeReactComponent used with ReactFake and iaux.IAReactComponent used with React should work the same. (e.g. ParentTileImg works with both)
+    constructor(props) {
+        super(props);
+        this.state = {}; // React doesnt do this
+        // In both React & ReactFake, _isMounted is set to true on loading
+        this._isMounted = false;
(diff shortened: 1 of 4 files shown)


mitra42:
@iisa as discussed, here's the PR for first 3 components

packages/ia-components/sandbox/IAReactComponent.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+const debug = require('debug')('dweb-archive:ConfigDetailsComponent');
```
iisa: Suggestion: Add comment around `debug` component - and how to configure.  You mentioned that it is a current node process but there is a client side workaround to ensure that it works.

mitra42:
Hmmm - that would put the comment on EVERY file, which seems syntactic overkill. 
The trick is: Server-side: 
`process.env.DEBUG="dweb-archive:*"` or `process.env.DEBUG="dweb-archive:ConfigDetailsComponent"` in the main JS file (e.g. in index.js if tht is what is requiring all these)

Client side it is:   `<script type="text/javascript">localStorage.debug = "dweb-archive dweb-archive:* dweb-transports dweb-transports:* dweb-objects dweb-objects:*";</script>`  or can execute that in the javascript before the `require debug` 

If neither are done, the debug doesnt slow things down , I forget if it just compiles to a noop or effectively that.

iisa:
Ok, that is fine for now.
For the record, we want the site/components to work without localStorage overall.  We will kick that can down the road, and address it when archive.org touches the component.

mitra42:
Understood - and note the client-side localStorage is only required to turn the debugging on, so could be easily commented out for normal use, or I'll figure out how to enable it in some other way when the need arises.

mitra42:
I'm working on those components at the moment - getting rid of the dweb dependencies, but will do that as a separate PR

PR #61, opened 2019-04-03
Adds -p flag to create directory only if it does not exist

### LINKED ISSUE #21: Jest tests failing on second time running storybook

shaneriley:

Closes #21 

Adds -p flag to skip creating the directory if it already exists.

From the "packages/ia-components" directory, run `yarn run storybook`, stop the running process, then run it again.

mkdir_p

### DIFF
--- a/packages/ia-components/package.json
+++ b/packages/ia-components/package.json
@@ -6,7 +6,7 @@
     "scripts": {
         "storybook": "yarn run test:generate-output && yarn run storybook:start",
         "storybook:start": "start-storybook -p 9001 -c .storybook",
-        "test:generate-output": "mkdir jest-test-utils && jest --json --outputFile=jest-test-utils/jest-test-results.json || true",
+        "test:generate-output": "mkdir -p jest-test-utils && jest --json --outputFile=jest-test-utils/jest-test-results.json || true",
         "test:update-snapshots": "jest --updateSnapshot",
         "test": "jest --watch --coverage"
     },


packages/ia-components/package.json
```diff
@@ -6,7 +6,7 @@
     "scripts": {
         "storybook": "yarn run test:generate-output && yarn run storybook:start",
         "storybook:start": "start-storybook -p 9001 -c .storybook",
-        "test:generate-output": "mkdir jest-test-utils && jest --json --outputFile=jest-test-utils/jest-test-results.json || true",
+        "test:generate-output": "mkdir -p jest-test-utils && jest --json --outputFile=jest-test-utils/jest-test-results.json || true",
```
mc2: `-p` makes this fail in Windows
`|| true` makes this fail in Windows

shreyansh23: @mc2 Do we expect that Windows users to use powershell or cmd or bash emulator?

mc2: cmd.exe, typically, LCD: lowest common denominator

mc2: interesting: https://www.npmjs.com/package/run-script-os

But maybe this directory is best pre-created.

shaneriley: @mc2 are you on a Windows environment where you could test some code in the command prompt? I'm wondering if we can just rely on `fs` to check for its existence, and then create it if not, like `!fs.existsSync(directory) && fs.mkdirSync(directory);`

iisa:
adding this here as part of possible solution sets: https://github.com/internetarchive/iaux/pull/23

iisa:
any thoughts on doing this? i am open to suggestions, refactor, etc.

PR #86, opened 2019-04-11
Port from Dweb: Fix TileComponent and add AnchorDetails

mitra42:
Remove (most) dweb dependencies from TileComponent, and adds AnchorDetails

There are still a couple of fields in TileComponent that are available in Dweb, but its unclear where they come from in a server-side implementation, especially the title of the first collection, since that is required for the text on the Parent-Collection popup image. 

Its tested in dweb-archive using both React or FakeReact, and will be pushed to live version soon. I don't have an obvious way to test in IAUX itself. 

### DIFF
--- a/packages/ia-components/sandbox/tiles/TileComponent.js
+++ b/packages/ia-components/sandbox/tiles/TileComponent.js
@@ -5,16 +5,27 @@ import IAReactComponent from '../IAReactComponent'; // Encapsulates differences 
 //import PropTypes from 'prop-types'
 //TODO-IAUX need to standardise API as this uses the "ArchiveMemberSearch" class to provide necessary details for the Tile.
 //import ArchiveMemberSearch from "@internetarchive/dweb-archivecontroller/ArchiveMemberSearch";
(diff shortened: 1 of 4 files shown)


packages/ia-components/sandbox/tiles/TileComponent.js
```diff
@@ -5,16 +5,27 @@ import IAReactComponent from '../IAReactComponent'; // Encapsulates differences
...
-import Util from "../../Util"; //TODO-IAUX for mediatype_canonical and number_format which need porting to some kind of common libary between IAUX and dweb-archive
 import TileImage from "./TileImage";
 import ParentTileImg from "./ParentTileImg";
-//Both these fail - not compiling the JSX
-//import TileImage from "@internetarchive/ia-components/sandbox/tiles/TileImage";
-//import TileImage from "../../iaux/packages/ia-components/sandbox/tiles/TileImage";
+import AnchorDetails from "../AnchorDetails";
+
+
+function number_format(nStr) //this is just addCommas now
```
iisa: love a good helper function. will keep note of this to extract in the future when we have a place for them to be used for all

PR #91, opened 2019-04-13
Fix extra unneccessary Promises , and document some of the places whe…

### LINKED ISSUE #90: Clean up promise structure
**Description**
Promise structure left over from original IAUX is odd, it makes many of the same mistakes I did when starting with promises, I fixed it in the dweb branch, but that never got merged, and the code has morphed too much to apply.  
At the meeting with @bfalling  & @iisa  I was asked to resubmit. 

This is pending (and will be a priority for me) once #86 is merged as given the experience twice already I don't want to have a stack of PR's that get out of date again before they are merged.

mitra42:
> **Description**: Fixes poor promise architecture, extra wrapping promises removes, including some async functions that return a promise in a promise in a promise ! Also documents some (not all) of the places where network functions are likely to cause an error, but the code isn't checking for that error or documenting that the functions themselves could fail.  

Closes #90 

a) its nonsense to reject(data), it should always be reject(err)
b) none of the code actually checked for those rejects coming back
c) async automaticlly returns a promise, no need to wrap in another one
d) fetch.then.catch also returns a promise so no need to wrap that in a promise or async
e) inside async should return or throw, not reject or resolve. 
f) anything that can throw an error (from network functions, which always have failure cases) should IMHO document that fact, or catch the error. 

I don't have a good way to test this as my code doesnt use these calls at all yet (more cleanup required before that can happen)

I don't have a good way to test this as my code doesnt use these calls at all yet (more cleanup required before that can happen). If there are some tests here and they were 
a) working on old code
b) failing on new code (this PR)
Please let me know how to run the test and I'll fix the code!

### DIFF
--- a/packages/ia-components/sandbox/iajs-test-component/index.jsx
+++ b/packages/ia-components/sandbox/iajs-test-component/index.jsx
@@ -31,6 +31,7 @@ export default class extends Component {
 
   fetchMetadata(identifier) {
     let metadataService = new MetadataService()
+    // TODO-ERRORS Note this has an unhandled error if MetadataService.get fails
     metadataService.get({identifier: identifier}).then(metadata => {
       this.setState({metadata: metadata.data})
     })
--- a/packages/ia-components/sandbox/related-as-list.js
+++ b/packages/ia-components/sandbox/related-as-list.js
@@ -18,6 +18,7 @@ export default class extends React.Component {
     }
 
     const relatedService = new RelatedService()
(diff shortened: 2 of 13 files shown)


iisa:
Promise clean up, reducing nested promise wrappers, and will defer reject/resolve to top layer that is using the class.  Will merge in for now, and make updates accordingly, so we can move forward with changes.  Tests still need to be done, but will test once we start using it.

PR #99, opened 2019-04-17
Add fallback cases for music player - no background photo, no tracks

iisa:
Add fallback cases for music player - no background photo, no tracks + unit tests


iisa:
thanks for testing @koderjoker. yes let's break that out to its own issue, great idea!

PR #101, opened 2019-04-18
changed css properties of tracks when hovered

Batey96:


packages/ia-components/sandbox/theatres/components/track-list/track-list.less
```diff
@@ -82,4 +82,10 @@
       float: right;
     }
   }
+  .track:hover{
```
iisa: hi @Batey96 , what brought this update on?

PR #102, opened 2019-04-19
Console.log > Debug

mitra42:
**Description**
Moves from console.log to debug
**Technical**
debug has slightly different syntax, its a format string like ("foo = %s", foo) though there are no examples in the code I changed. See https://www.npmjs.com/package/debug 
it need a line like `const debug = require('debug')('ia-components:thiscomponent');`
Also ... debugging specfic components or the repo are controlled by the line in `packages/ia-prototype-apps/apps/details-react/index.html`
 `<script type="text/javascript">localStorage.debug = "ia-components:* ia-js-client:*";</script>`
which has to occur before the require.  (There is a similar control line if this is used in node). 

**Testing**
Again ... unclear that there are any tests in IAUX to run yet.


packages/ia-prototype-apps/apps/details-react/index.html
```diff
@@ -12,6 +12,8 @@
     <script src="https://archive.org/bookreader/BookReaderHelpers.js"></script>
     <script src="https://archive.org/bookreader/LendingFlow.js"></script>
     <script src="https://archive.org/bookreader/BookReaderJSIA.js"></script>
+    <-- turn on debugging, comment out for no debugging-->
```
iisa: @mitra42 is this where you were able to successfully test?

packages/ia-js-client/src/controllers/item/item.ts
```diff
@@ -1,5 +1,6 @@
 import { MetadataService, Metadata } from '../../services/metadata'
 import { DetailsService, DetailsResponse } from '../../services/details'
+const debug = require('debug')('ia-js-client:item');
```
iisa: I remember you mentioning that this is an npm module? I can't recall..
Is there a problem with using import statement?

packages/ia-components/sandbox/AnchorDetails.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import IAReactComponent from './IAReactComponent';
-const debug = require('debug')('dweb-archive:AnchorDetails');
+const debug = require('debug')('ia-components:AnchorDetails');
```
iisa: can we use import statement?
how do we turn off when we use in prod?
maybe we need an entry in the README with a little bit of usage direction

packages/ia-prototype-apps/apps/details-react/index.html
mitra42: Sorry - just saw this comment after the call, no I have currently no way to test anything in IAUX, I'm more than happy to do so but not sure that we have anything there that is testable standalone.   There is a PR proposed yesterday on dweb-archive to allow that to use components from ia-components, if that works I'll be actively testing ia-components & using in my (live) code from there on.

packages/ia-js-client/src/controllers/item/item.ts
mitra42: I haven't ever seen debug used with import, I think import is async, and I'm pretty sure it has to be loaded before the javascript code (I know the environment variable or localstore has to be set before the "require debug" in order for debug to be enabled. I don't quite understand why but think its to minimise the impact on production code in the case that debugging is off.

Issue #108, opened 2019-04-20
Setup fails due to erros in ia-js-client package


Batey96:
**Description**
ON a fresh clone and install,errors are being encountered when bootstrapping the packages.This is as a result of typo errors in the ia-js-client package..

**Steps To Reproduce**
```
git clone https://github.com/internetarchive/iaux.git
cd "cloned Repo"
yarn install 
yarn run lerna bootstrap
```

![iajsfail](https://user-images.githubusercontent.com/24868289/56456896-aeb08d00-636a-11e9-996b-60f9020ed0ae.png)

mitra42:
Are any of them a result of the changes I made recently ? 

I need a scroll-up on your screen shot, as I can't see what you are running that generates these errors, can you post the STR and I'll check the bugs aren't something I introduced.

Batey96:
Yes i think so.the recent changes made the bug!..because i compared it with an older ia-js-client package and bootstrapping was okay..the additions you made handling fallbacks are a source of it..typo errors which i fixed.PR sent

Batey96:
PR 109

mitra42:
Thanks - I posted comments in

Batey96:
Steps To Reproduce
```
git clone https://github.com/internetarchive/iaux.git
cd  "cloned Repo"
yarn install
yarn run lerna bootstrap
```

mitra42:
Thanks @Batey96 for adding them above.

Batey96:
Issue resolved

PR #109, opened 2019-04-20
Fixed setup errors due to typo in related.ts and item.ts(issue #108)

Batey96:
**Description**
After this fix,all packages installed succesffully..

**Evidence**
fixediajsclient

### DIFF
--- a/packages/ia-js-client/src/controllers/item/item.ts
+++ b/packages/ia-js-client/src/controllers/item/item.ts
@@ -5,32 +5,32 @@ import { DetailsService, DetailsResponse } from '../../services/details'
  * early prototype of audio file class (will change)
  */
 export class AudioFile {
-  name:string
-  length:number
-  track:string
-  artist:string
-  sources:Array<{
-    file:string
-    type:string // eg mp3
-    height:string|number
-    width:string|number
+  name: string
+  length: number
+  track: string
+  artist: string
+  sources: Array<{
+    file: string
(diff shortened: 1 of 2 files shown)


mitra42:
@Batey96 thanks for catching this and sorry I let them slip through.

A few minor points to improve the flow. 

a) If there is a related issue, its helpful to reference it - in this case its #108 , otherwise @iisa isn't going to know which bug this fixes. 

b) both in #108 and here, the STR are missing, i.e. steps where someone else can see the bug, and neither of your screen shots show what commands you were running to trigger the error. I'm sure its obvious to you, but its not to me since I'm not using the ia-js-client code yet.  Could you post those STR here and on #108. 

c) there is major linting in this PR which makes it hard to find the 2 one-line changes you made. IMHO it would be much better to post significant linting (adjustment of whitespace) as a separate PR explicitly marked (ONLY LINTING). In general major linting also makes it really hard for any other changes made, by someone else, at the same time to get merged since every line will flag a conflict.
  
I'm not sure how/if automatic linting is setup in this repo - I see .eslintrc, but I'm not familiar enough with it to see if its enforcing spacing. if that is just your own (perfectly reasonable) personal preferences in your IDE, could you turn it off as it will make it hard for anything you change to get merged with anything someone else changes - the spacing in those files came from Richard and while its a little odd for my preference as well, I intentionally didn't change it as it makes everyone else's work harder to do so.

PR #115, opened 2019-04-23
Adding localhost proxy to StorybookJS so we can make API calls

iisa:
Problem:
Making AJAX calls inside of a StorybookJS component will cause a CORS error

Solution:
Piggyback on StorybookJS' dev server and route external calls through a proxy


packages/ia-components/.storybook/middleware.js
```diff
@@ -0,0 +1,37 @@
+const proxy = require('http-proxy-middleware');
```
shaneriley: I'd refactor this file to make it easier to both change the base URL for the API (in the event of a dev or staging env) and to quickly add routes, like:

```
const base_url = "https://archive.org";

const constructOptions = (route) => (
  {
    target: `${base_url}/${route}`,
    logLevel: 'debug',
    changeOrigin: true,
    pathRewrite: {
      `^/${route}`: `${base_url}/${route}`
    }
  }
);

module.exports = function expressMiddleware (router) {
  ["download", "images", "serve"].forEach((route) => {
    router.use(`/${route}`, proxy(constructOptions(route)))
  });

  router.use('/api', proxy({
    target: base_url,
    logLevel: 'debug',
    changeOrigin: true,
    pathRewrite: {
      '^/api': base_url
    }
  }));
};
```

packages/ia-components/package.json
```diff
@@ -21,8 +21,10 @@
         "@storybook/addon-knobs": "^4.1.4",
         "@storybook/react": "^4.1.4",
         "assert": "^1.4.1",
+        "axios": "^0.18.0",
```
shaneriley: I don't see Axios being used in this commit.

packages/ia-components/.storybook/middleware.js
mc2: Great contribution, Shane. Agree.

iisa: that's a good optimization, will apply it.

PR #122, opened 2019-04-25
Track List story

sarru1291:
**Description**

> What does this PR achieve?
User interface story for the only tracklist.
Screenshot from 2019-04-25 09-22-16

### DIFF
--- a/packages/ia-components/sandbox/theatres/components/track-list/track-list.stories.js
+++ b/packages/ia-components/sandbox/theatres/components/track-list/track-list.stories.js
@@ -5,7 +5,6 @@ import { text, number } from '@storybook/addon-knobs';
 import TrackList from './track-list';
 
  storiesOf('Theatres', module)
-  // .addParameters({ jest: ['example.test.js'] })
   .addWithJSX('Tract List', () => {
     const tracks = [
       {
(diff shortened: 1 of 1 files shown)


Batey96:
Great but i think what @isa wants is for you to maintain the data flow from the current code strucuture..
There is already a tracklist component..the goal here is to have a choice to use the youtube component over the current embedded one..All things being equal

sarru1291:
> Great but i think what @isa wants is for you to maintain the data flow from the current code strucuture..
> There is already a tracklist component..the goal here is to have a choice to use the youtube component over the current embedded one..All things being equal

I was testing the tracklist that's why I made its story. Currently, I'm about to complete the youtube component.

packages/ia-components/sandbox/theatres/components/track-list/track-list.stories.js
```diff
@@ -1 +1,40 @@
 import style from './track-list.less';
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { text, number } from '@storybook/addon-knobs';
+import TrackList from './track-list';
+
+ storiesOf('Theatres', module)
+  // .addParameters({ jest: ['example.test.js'] })
```
iisa: let's remove this

sarru1291:
it's resolved now.

PR #132, opened 2019-05-02
Add details components including Tabby 

mitra42:
**Description**
I've added a "details" folder for the small components that make up the parts of the detail page. 

In it, I've put my first - which is a single Tabby component (a set of Tabbys provide the UI for switching between tabs, but they vary significantly between uses so I haven't componentized them yet. 

**Technical**
Its dependent on code in archive.js, which could usefully in the future be moved into this component. The archive.js code is tangled enough that I haven't touched it except when I need to change something. 

**Testing**
Not tested in IAUX but this component is in use in dweb-archive

Note - as pushed, it builds on #131 (Readme for debug), I think once that is merged it should automatically be seen as just a single file and new directory.


mitra42:
I've added a number of other small components to that directory that I've been using on dweb-archive, together they form a lot of the parts of the details page that aren't about displaying content

iisa:
looks good. we will update file extensions and lint as a follow.

PR #142, opened 2019-05-05
Switch all new components to jsx and include in index.js

mitra42:
**Description**
As discussed with @iisa  - switched all new components from dweb-archive to jsx and to import via index.js

**Technical**
Unfortunately webpack doesnt understand that .jsx is an automatic extension so had to expicitly specify that in some places. 

Also index.js imports IAUXExampleComponent which uses an experimental extension classProperties. Webpack from other packages breaks because it doesnt understand this.  
Rather than have every other package figure out how to enable this, (which looks non-trivial) I suggest that IAUXExampleComponent be imported directly.  

**Testing**
Still no understanding of any tests to run ... 
Its tested in dweb-archive#iaux, which will probably be merged into dweb-archive#master soon since its looking good.

**Evidence**
N/A


mitra42:
Isa - we may want to discuss whether to disable IAUXExampleComponent in index.js OR document adding classProperties

Either way ... I don't think examples should be in index.js as I think it will increase the bundle size for the result ?

packages/ia-components/index.js
```diff
@@ -3,7 +3,10 @@
  */
 
 // live
-export { default as IAUXExampleComponent } from './live/example-react-component/index';
+//Next line fails webpack - complains Support for the experimental syntax 'classProperties' isn't currently enabled
```
iisa: @mitra42 can you send me the webpack config that you are using? 

you may need to add `.jsx` into your bundling config.  i can help debug, that way we can drop the `.jsx` from the import statements.

mitra42: I have https://github.com/internetarchive/dweb-archive/blob/master/.babelrc and https://github.com/internetarchive/dweb-archive/blob/master/webpack.config.js 
edited recently so it does the webpack by @shreyansh23 and then I added the `js[x]`. 

The remaining problem is that for some reason it wasn't getting the import of DetailsFlags from DetailsActionButtons without it being DetailsFlags.jsx. I'll try and repeat the failing scenario to remind myself.

mitra42:
I wasnt able to repeat the problem with imports within ia-components without specifying .jsx, I pushed a change to that PR to remove those .jsx extensions.  So all 3 PRs are ready for merging (from my perspective).

packages/ia-components/index.js
```diff
@@ -3,7 +3,10 @@
...
+//If it is needed then please document how to add it to something using this index.js from another package
+//or import directory
+//export { default as IAUXExampleComponent } from './live/example-react-component/index';
```
iisa: this may break global AJS archive.org. it is importing it

packages/ia-components/sandbox/details/AnchorDownload.jsx
```diff
@@ -3,49 +3,60 @@ import IAReactComponent from '../IAReactComponent';
...
-    <AnchorDownload
-        identifier  of item
-        filename    Name of file to download
-        source      [ArchiveFile]
-        format      argument to formats parameter - if this is present we'll download the .../compress/IDENTIFIER/formats=FORMAT
-
-        Any other parms are passed to the anchor
-    />
+/**
+ * Component used as an anchor where it should engage part of the download functionality.
```
iisa: 👍

mitra42:
Thanks - and worth considering the issue raised 
a) should we be requiring future enhancements in any of  these releases - I'd be a bit concerned about those working everywhere, but maybe someone has done the research on which ones are stable
b) should index.js be including examples - I don't think so, I think examples should either have their own index or more likely be separated included

PR #145, opened 2019-05-08
Paginator: keep page scroll fixed when autofocusing item

iisa:
**Description**
When paginator is on a page, it will autofocus an item.  We should not be jumping to the autofocus.

**Technical**
When `element.focus()` is called, the browser will automatically jump to make that element viewable.
We need to preserve the User's scroll position.


iisa:
internal QA looks good, merging.

PR #148, opened 2019-05-14
Adds appearance Less mixin

shaneriley:
**Description**

Adds a Less mixin for the `appearance` property.

**Technical**

Accepts one required argument, the value of the property.

**Testing**

To be completed in Petabox with ticket WEBDEV-2221.

**Evidence**

N/A


mekarpeles:
lgtm, could/should we be auto-adding prefixes with some tool?

shaneriley:
We could, but I didn't want to go so heavy handed yet since it will further slow the compiling of assets on review apps during development.

PR #149, opened 2019-05-15
Music Player, JWPlayer wrapper: manually play the track that the player starts on

iisa:
UX Problem:
The user cannot select highlighted track on track list to start playing track when player doesn't start on first track.

Technical Problem:
This regression was introduced when Play8.js was updated to handle new browser issues that affected overal UX.
This meant that the initial load of Play8 with JWPlayer has changed from 1 to 2 callback emits. This is a known count across browsers.

Technical Solution:
- Add the notion of when JWPlayer load has stabilized when player starts at N (where N > track 1).  This needs:
  - manual counts of how many times JWPlayer fires `onPlaylistItem` event on load
  - manual checks when state updates in `shouldComponentUpdate` and `componentDidUpdate`
  - manual calls to global jwplayer outside of Play8 wrapper

This update did not affect what currently works: autoplay to next track, cold loads on track 1, clicking on other tracks after load

Tested on Chrome, Firefox, Safari
WEBDEV-2236

### DIFF
--- a/packages/ia-components/sandbox/theatres/components/audio-player/players_by_type/archive-audio-jwplayer-wrapper.jsx
+++ b/packages/ia-components/sandbox/theatres/components/audio-player/players_by_type/archive-audio-jwplayer-wrapper.jsx
@@ -3,10 +3,12 @@ import PropTypes from 'prop-types';
 
 /**
  * IA Audio Player
- *
- * Uses global: Play class (IA JWPlayer wrapper)
+ * Uses global: Play class (IA JWPlayer wrapper), jwplayer
  *
  * It will display photo if given, and will overlay the media player at the base of the photo.
+ *
+ * Development is in lock-step with Play8.js.
+ * ANY changes to Play8.js may need to be addressed here also.
  */
 class ArchiveAudioPlayer extends Component {
   constructor(props) {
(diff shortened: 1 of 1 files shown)


packages/ia-components/sandbox/theatres/components/audio-player/players_by_type/archive-audio-jwplayer-wrapper.jsx
```diff
@@ -32,51 +40,139 @@ class ArchiveAudioPlayer extends Component {
...
+    if (jwplayerStartingOnACertainTrack && !initiatePlay) {
+      if (playCBHasFired === 0) {
+        // capture the index where the playlist starts & start count of CB fires
+        this.setState({ startAt: playerPlaylistIndex, playCBHasFired: playCBHasFired + 1 });
+      }
+
+      if (playCBHasFired === 1) {
+        // Once CB has fired once, we will continue to count the CB fire
+        // & emit that the player has updated index
+        this.setState({ playCBHasFired: playCBHasFired + 1 },
```
shaneriley: Since this is only setting if `playCBHasFired` is 1, can we just set the new value to `2`?

iisa: good suggestion. i will prefer to leave it as `+ 1` shows that we are counting.

PR #150, opened 2019-05-16
(Partial) Fix dependency problem on dweb-archivecontroller

mitra42:
**Description**

The components pushed to iaux from dweb-archive use the more functional controllers in dweb-archivecontroller, BUT this now creates a pack-time dependency on dweb-archivecontroller.

**Technical**

Some of that dependency is function that could useful go in util.js, this PR fixes that

**Testing**
@iisa to try repacking on petabox, I have no way to test that

**Evidence**


mitra42:
@iisa - this was for the problem you told me with https://git.archive.org/ia/petabox/-/jobs/152663

We can discuss the other one tomorrow.

PR #151, opened 2019-05-16
Example Component: readding into index.js, removing experimental `classProperties` usage

iisa:
Petabox builds are failing without this import.
Imports are failing DWEB usage of IAUX.
To simplify, we are removing this line.


iisa:
@mitra42 has approved

PR #154, opened 2019-05-17
Adding @internetarchive/dweb-archivecontroller

iisa:
**Description**
Dweb components rely on this controller import.

In order for build to pass in Petabox, this needs to be included.  I have conferred with @mitra42 and the tag is the most current for this import.

I've tested local Petabox container build and is finally successful with this update.

### DIFF
--- a/packages/ia-components/package.json
+++ b/packages/ia-components/package.json
@@ -33,6 +33,7 @@
         "style-loader": "^0.23.1"
     },
     "dependencies": {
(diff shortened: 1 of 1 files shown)


bfalling:
LGTM

mitra42:
@iisa Does merging this remove the need for the hack in https://github.com/internetarchive/iaux/pull/155#event-2350346639 ?

PR #155, opened 2019-05-17
Fix dweb dependency

iisa:
Recent integration of `dweb-controller` is breaking PETABOX build.

This fix:
- Lints `DetailsCollectionList` - [commit](https://github.com/internetarchive/iaux/commit/334d9df8fb2593740f7dbf3dab32ffdba43af39c)
- removes the `dweb-controller` dependency
- refactored `DetailsCollectionList` to remove dependency & to accept a new prop `ArchiveMemberExpand` - [commit](https://github.com/internetarchive/iaux/commit/ce1ba69f589fe63e56f39c4a28cfed100843b3db)

Note: Look at each commit to see individual changes

**Evidence**
Image of tests passing locally:
Screen Shot 2019-05-17 at 12 32 21 PM

### DIFF
--- a/packages/ia-components/package.json
+++ b/packages/ia-components/package.json
@@ -33,7 +33,6 @@
         "style-loader": "^0.23.1"
     },
     "dependencies": {
-        "@internetarchive/dweb-archivecontroller": "^0.1.57",
         "debug": "^4.1.1",
         "lodash": "^4.17.11",
         "prop-types": "^15.6.2",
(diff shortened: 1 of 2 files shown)


iisa:
this is part of https://github.com/internetarchive/dweb-archive/issues/101 - DWEB needs to follow up with this fix - to pass `ArchiveMemberExpand` function in as a prop.

iisa:
@jbuckner has approved via side-by-side code review

mitra42:
Yuck - this is a horrible fix :-) because its setting up a bad pattern of the components not being able to access any of the Controller functions unless they are passed around in the HTML as props.   This is just one example of a component that will need to access the controller.

iisa:
this is our solution: https://github.com/internetarchive/iaux/pull/156

PR #156, opened 2019-05-17
Removing DetailsCollectionsList.jsx from IAUX

iisa:
- Currently, the component's external query side effect is breaking Petabox build
- This also is an anti-pattern of IAUX's premise - encapsulated display components
- Component will be reworked in dweb side and will be MR'd when ready.

Build compiles and tests passing
image


iisa:
Per @mitra42 and I's conversation around component responsibility, this is the best solution to make at this time.

PR #168, opened 2019-05-27
Mitra uxlocal - interface for managing local resources

mitra42:
**Description**

This provides support for managing locally downloaded resources in dweb-mirror - it adds an indicator on Tiles when the item is downloaded. 

**Technical**

Its dependent on most of the other PR's I've already submitted, actual changes are only 
Adding CrawlConfig.js; and patching into NavWrap and TileComponent

**Testing**
Live in dweb-mirror

**Evidence**


packages/ia-components/sandbox/details/CrawlConfig.js
```diff
@@ -0,0 +1,145 @@
+import ReactDOM from "react-dom";
+
+const debug = require('debug')('dweb-archive:CrawlConfig');
+const canonicaljson = require('@stratumn/canonicaljson');
```
iisa: are these able to be imports?  or do they need require?

mitra42: debug definately needs require (it has to be loaded before the debug statement) 
canonicaljson probably doesn't ... I'll make that change.

mitra42: Done

iisa:
@mitra ^^ conflicts.

iisa:
@mitra, the commit history is not reflective of changes here as well. i would like to get this in via squashed merge but am unsure of how this will affect your other branches. perhaps we should coordinate merges

cc @shaneriley

mitra42:
Most of that commit history is merges i.e. rebases, done, partly at your request, IN ORDER to make the actual difference simpler. but if you look at https://github.com/internetarchive/iaux/pull/168/files there are only 3 file changes, (an addition of CrawlConfig.jsx, a change to NavWrap to use it (the one you found in #163 on its own) and the small UI change that this PR is all about.

PR #203, opened 2019-06-21
Jwplayer round3

iisa:
**Description**
 
- Play8 `onSetupComplete` sends custom callback that signals what track jwplayer starts on
- only fires `PlayN` when user selects track or when track auto advances
- keeps player element on page so it matches the `Play` instance that it creates on first mount for the whole session


iisa:
went through this side by side with @traceypooh . we are going with it.

PR #206, opened 2019-06-21
Creating DWEB index file

### LINKED ISSUE #205: Create a new index file for dweb
DWeb components are coming in hot and we need to ensure that they are able to be received and used by dweb at any given time.

Meanwhile, we should introduce them into the main index file once we start working on em

iisa:
**Description**
creates new dweb index.
Closes #205


mitra42:
Looks good - I'll need to check/change the import's but that's trivial and at my next webpack stage

PR #208, opened 2019-06-23
Replace YouTube embed with YouTube API integration

### LINKED ISSUE #178: Replace YouTube embed with YouTube API integration
**Context**

Our integration with YouTube’s player is currently very limited. We’re basically just using their default embed. We can respond a lot better by using their API.

**Proposal & Constraints**

Replace the basic embed with integration with their API, which will provide the following benefits:

Ability to detect when song’s end, in order to play next

Ability to detect when there is a problem with a video, so we can display alternate messaging

Keep in mind that YouTube restricts too heavy use of their API’s.

An example of an album with lots of broken YouTube id’s: https://archive.org/details/cd_dark-side-of-the-moon_pink-floyd/disc1/02.+Pink+Floyd+-+On+the+Run.flac

@iisa will help define sub-tickets

koderjoker:
**Description**

Closes #178 

**Task List**

- [x] To replace basic embedded with Youtube iframe API
- [x] To play video of any track selected by a user
- [x] To autoplay next track when video finishes playing
- [x] To move on to the next track if player encounters an error
- [x] To avoid restarting video if the same track is selected
- [x] To ensure tracklist is rendered in all instances
- [x] To loop back to first track at the end of playlist
- [x] To prevent video loading till the player is ready
- [x] To ensure player is rendered again after switching through channels
- [x] To ensure player works on mobile devices
- [x] To remove comments 
- [x] To add jsdocs
- [x] To make sure files are in the correct location
- [x] To delete redundant code
- [x] To ensure code passes linter check
- [x] To rebase to include latest changes

**Evidence**

Track selection
Youtube API integration

Autoplay
Autoplay

Error handling
Error handling

### DIFF
--- a/packages/ia-components/sandbox/audio-players/youtube-player.jsx
+++ b/packages/ia-components/sandbox/audio-players/youtube-player.jsx
@@ -1,8 +1,13 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
-// No restrictions for using iframe :)
-// functions and parameters such as autoplay, playVideo(), loadVideoById() won't work in all mobile environments
+/**
+ * YoutubePlayer
+ *
+ * Loads the youtube player and plays video of the selected track
+ *
+ * @params see PropTypes
+ */
 
 class YoutubePlayer extends Component {
 
@@ -28,7 +33,10 @@ class YoutubePlayer extends Component {
     });
   }
 
-  // load api and iframe object
+  /**
+   * Load the youtube iframe API if not previously loaded
+   * Load the youtube player object when youtube iframe api is ready
+   */
   componentDidMount() {
     if (!window.YT) {
       this.loadAPI();
@@ -40,7 +48,10 @@ class YoutubePlayer extends Component {
     }
   }
 
-  // Re render component only if track number changes and iframe has loaded
+  /**
+   * Update component only if there is a change in video id
+   * If above update video id and selected track
+   */
   shouldComponentUpdate(nextProps) {
     const { id } = this.state;
     const trackChanged = id !== nextProps.id;
@@ -53,13 +64,18 @@ class YoutubePlayer extends Component {
     return trackChanged;
   }
 
-  // if component updates load video with received id
+  /**
+   * Load video of passed id and default resolution
+   */
   componentDidUpdate() {
     const { id, player } = this.state;
-    player.loadVideoById(id, 'default'); // change resolution?
+    player.loadVideoById(id, 'default');
   }
 
-  // Load iframe API async
+  /**
+   * Load youtube iframe API asyncronously
+   * Prevent user interaction with tracklist
+   */
   loadAPI() {
     document.querySelector('.audio-track-list').setAttribute('style', 'pointer-events: none');
     const tag = document.createElement('script');
@@ -68,7 +84,10 @@ class YoutubePlayer extends Component {
     firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
   }
 
-  // Load iframe object
+  /**
+   * Load youtube player object
+   * Update player
+   */
   loadPlayer() {
     const { id, playerAnchor } = this.state;
     this.setState({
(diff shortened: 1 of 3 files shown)


iisa:
@shaneriley, I tagged you as a reviewer for @koderjoker's code. She can benefit from your deep knowledge.  Currently, it's WIP, but she has requested an initial review for it so she can iterate.  Would you be able to check it out?

shaneriley:
Yeah, I can review it this evening.

packages/ia-components/sandbox/audio-players/audio-player-main.jsx
```diff
@@ -72,16 +73,31 @@ export default class TheatreAudioPlayer extends Component {
...
         {...this.props}
         onRegistrationComplete={this.receiveURLSetter}
         needsURLSettingAccess
       />
-    );
-    if (isExternal) {
-      // make iframe with URL
+    ); // shorten code
+    if (source === 'youtube') {
+      const { urlSetterFN } = this.state;
```
shaneriley: This is being extracted in both conditions and passed a `trackNumber` if it's defined. This definition, the definition of `trackNumber`, and the resulting conditional to call `urlSetterFN` can be pulled out of both of these parent conditionals and simply performed after both of them.

packages/ia-components/sandbox/audio-players/audio-player-main.jsx
```diff
@@ -72,16 +73,31 @@ export default class TheatreAudioPlayer extends Component {
...
+      const externalSourceDetails = sourceData[source] || {};
```
shaneriley: This variable instantiation is identical in both conditions, and can therefore be done prior to the conditional blocks.

packages/ia-components/sandbox/audio-players/audio-player.stories.js
```diff
@@ -1 +1,2 @@
 import style from './audio-player.less';
+import styles from './youtube-player.less'
```
shaneriley: We should rename these two variables to something more easily identifiable, like `audio_styles` and `youtube_styles`

packages/ia-components/sandbox/audio-players/youtube-player.jsx
```diff
@@ -0,0 +1,120 @@
...
+    const { selectedTrack, id } = props
+
+    this.state = {
+      playerAnchor: React.createRef(),
+      selectedTrack,
+      id
+    }
+
+    // use static methods?
+    this.loadAPI = this.loadAPI.bind(this);
```
shaneriley: This can be a class method instead, since it has no reliance on the class instance.

packages/ia-components/sandbox/audio-players/youtube-player.jsx
```diff
@@ -0,0 +1,120 @@
...
+    this.loadPlayer = this.loadPlayer.bind(this);
```
shaneriley: I would prefer to see methods bound to the instance's context using a loop of some sort, like:
```javascript
[
  'loadPlayer',
  'onPlayerStateChange',
  'onPlayerReady',
  'onPlayerError'
].forEach((method) => this[method] = this[method].bind(this));
```

packages/ia-components/sandbox/audio-players/youtube-player.jsx
```diff
@@ -0,0 +1,120 @@
...
+  shouldComponentUpdate(nextProps) {
+    const { id } = this.state;
+    if (id !== nextProps.id) {
+      this.setState({
+        id: nextProps.id,
+        selectedTrack: nextProps.selectedTrack
+      });
+      return true;
+    }
+    return false;
```
shaneriley: Rather than have two return statements in this method, we could simply `return id !== nextProps.id`, which I would do by way of a local `const` assignment to ensure the conditional code is updatable in one location. Something like: `const track_changed = id !== nextProps.id`

packages/ia-components/sandbox/audio-players/youtube-player.less
```diff
@@ -0,0 +1,14 @@
+.YoutubePlayer {
+  width: 99.5%;
+  height: 4.4rem; // accomodate height for jwplayer controls
+}
+
+.YoutubePlayer iframe {
+    position: absolute;
+    top: 0.1rem;
+    left: 0.1rem;
+    width: 100%;
```
shaneriley: When you're setting a value for all four positioning properties, width and height don't have an effect on the element and can be omitted.

packages/ia-components/sandbox/details/NavWrap.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import IAReactComponent from '../IAReactComponent';
 import AnchorDetails from "../AnchorDetails";
-import ConfigDetailsComponent from './ConfigDetailsComponent';
+// import ConfigDetailsComponent from './ConfigDetailsComponent';
```
shaneriley: Safe to remove?

packages/ia-components/sandbox/audio-players/audio-player-main.jsx
koderjoker: I'll be using `trackNumber` in the youtube player so I'll take the rest of the definitions and conditional out.

koderjoker: Additionally `urlSetterFN` also returns a value for archive and executes (and returns an error) for it when it's outside the youtube and spotify conditionals

packages/ia-components/sandbox/details/NavWrap.jsx
koderjoker: Resolved with rebase

packages/ia-components/sandbox/audio-players/youtube-player.less
koderjoker: Player expands out of the frame when height and width are omitted 😕

shaneriley: Odd. So does it already have width and height property or attribute assignments that you're overriding?

koderjoker: Yes, it's overriding the original height and width (600px) of the player.

PR #219, opened 2019-06-28
Fix scope issue in liner notes scaling

iisa:
Problem: The `this` in the `getPageURL` function was lost in Reactland.
Solution: pass in the `br` instance that was created

**Testing**

go to: https://www-isa.archive.org/details/cd_the-best-of-anita-baker_anita-baker-the-winans
and LP : https://www-isa.archive.org/details/lp_the-classic-jazz-quartet_the-classic-jazz-quartet

Liner notes should display.  This is especially for a fix for the LPs where the original sizes are massive.

**Evidence**
image

### DIFF
--- a/packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
+++ b/packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
@@ -20,6 +20,8 @@ export default class BookReaderWrapper extends Component {
 
   componentDidMount() {
     const { options } = this.props;
+
+    const originalGetPageURI = BookReader.prototype.getPageURI;
     const defaultOptions = {
       el: `#${this.BookReaderRef.current.id}`,
       mobileNavFullscreenOnly: true,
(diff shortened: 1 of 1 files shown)


packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
```diff
@@ -33,18 +35,13 @@ export default class BookReaderWrapper extends Component {
...
-      getPageURI: (index, reduce, rotate) => {
-        if ('undefined' == typeof(reduce))
-          reduce = 1;
-        if ('undefined' == typeof(rotate))
-          rotate = 0;
-        let uri = BookReader.prototype.getPageURI.call(this, index, reduce, rotate);
-        uri = uri + (uri.indexOf('?') > -1 ? '&' : '?');
-        uri = uri + 'scale=' + reduce + '&rotate=' + rotate;
+      getPageURI: (index, reduce = 1, rotate = 0) => {
+        let uri = originalGetPageURI.call(br, index, reduce, rotate);
```
bfalling: Question: Why use `br` instead of `this`?

iisa: `this` was commandeered by the React component once the function was called.  The actual instance we want in this case is the new BookReader instance which is tacked to the window as `window.br`.

PR #224, opened 2019-07-01
Clear timeout when new track selected

### LINKED ISSUE #223: settimeout and autoplay disrupt normal track progression
**Task Details**
After clicking Angel followed by Same Ole Love, tracks play unconsecutively because every track is trying to complete their settimeout and autoplay.
(Angel, You’re the Best Thing Yet, Same Ole Love and No One in the World are tracks with errors)

Jul-02-2019 01-27-50

koderjoker:
**Description**

Closes #223  

Clears timeout to prevent it from running when the next track is selected.

### DIFF
--- a/packages/ia-components/sandbox/audio-players/youtube-player.jsx
+++ b/packages/ia-components/sandbox/audio-players/youtube-player.jsx
@@ -15,6 +15,7 @@ class YoutubePlayer extends Component {
     super(props);
 
     const { selectedTrack, id } = props;
+    this.timer = null;
 
     this.state = {
       player: null,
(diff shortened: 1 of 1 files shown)


packages/ia-components/sandbox/audio-players/youtube-player.jsx
```diff
@@ -56,6 +56,7 @@ class YoutubePlayer extends Component {
     const { id } = this.state;
     const trackChanged = id !== nextProps.id;
     if (trackChanged) {
+      clearTimeout(window.timer);
```
shaneriley: Can we reference the timer on the component instead of the window? Then we're better safeguarded from some other code overwriting `window.timer`

koderjoker: Done!

PR #232, opened 2019-07-19
Check track before sending to youtube wrapper or spotify

### LINKED ISSUE #230: Stop youtube player from loading first track as default
**Task Details**

Currently, our youtube player is sent first track by default at the start. When the first track isn't available for the same, the youtube component is sent a blank id, and thus the creation of the youtube player fails.

The starting track needs to be checked if it has a video on youtube, before sending it to the youtube component. If not available, another succeeding track should be sent

koderjoker:
**Description**

Closes #230

Verifies a track number's presence in the youtube/spotify channel before sending it to their respective wrapper, if not present sends the next track in order.

### DIFF
--- a/packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify-main.jsx
+++ b/packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify-main.jsx
@@ -161,6 +161,7 @@ class AudioPlayerWithYoutubeSpotify extends Component {
     if (isAlbum) {
       audioSource = {};
       audioSource[channelToPlay] = albumSource;
+      return audioSource;
     }
     // If selected track not null find info for selcted track, else find for first track of list of selected channel
     audioSource = trackSelected ? tracklistToShow.find(t => t.trackNumber === trackSelected) : tracklistToShow.find(t => t.trackNumber >= 1);
--- a/packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify-main.jsx
+++ b/packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify-main.jsx
@@ -162,9 +162,8 @@ class AudioPlayerWithYoutubeSpotify extends Component {
       audioSource = {};
       audioSource[channelToPlay] = albumSource;
     }
-    const trackToHighlight = trackSelected === null && trackStartingPoint !== null ? trackStartingPoint : trackSelected;
-    audioSource = find(tracklistToShow, track => track.trackNumber === trackToHighlight);
-
+    // If selected track not null find info for selcted track, else find for first track of list of selected channel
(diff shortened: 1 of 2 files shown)


packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify-main.jsx
```diff
@@ -162,9 +162,8 @@ class AudioPlayerWithYoutubeSpotify extends Component {
       audioSource = {};
       audioSource[channelToPlay] = albumSource;
     }
-    const trackToHighlight = trackSelected === null && trackStartingPoint !== null ? trackStartingPoint : trackSelected;
-    audioSource = find(tracklistToShow, track => track.trackNumber === trackToHighlight);
-
+    // If selected track not null find info for selcted track, else find for first track of list of selected channel
+    audioSource = trackSelected ? tracklistToShow.find(t => t.trackNumber === trackSelected) : tracklistToShow.find(t => t.trackNumber >= 1);
```
shaneriley: I'm curious what happens when a CD is presented with a track number of 0. This album, for example, ripped the media before the first track marker as "Hidden track one audio". With this change wouldn't the player always skip that 0th track?

https://archive.org/details/cd_burn_sister-machine-gun

koderjoker: In this case track 0 is actually assigned the trackNumber 1 (and we're only depending on trackNumber to decide the track to be loaded)

track 0
Screen Shot 2019-07-19 at 3 06 00 PM

track 1
Screen Shot 2019-07-19 at 3 06 10 PM

shaneriley: 👍

packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify-main.jsx
```diff
@@ -162,9 +162,8 @@ class AudioPlayerWithYoutubeSpotify extends Component {
       audioSource = {};
       audioSource[channelToPlay] = albumSource;
     }
-    const trackToHighlight = trackSelected === null && trackStartingPoint !== null ? trackStartingPoint : trackSelected;
```
iisa: @koderjoker what happens to `trackToHighlight`?

koderjoker: The value of trackToHighlight (which is only sent to the TheatreTrackList) isn't affected since it's calculated in render.

iisa: what about line `163` where it is setting `audioSource` - should we return early there so this next line doesn't mutate the value?

PR #236, opened 2019-07-23
Music player components cleanup

iisa:
- added displayName field for easier debugging
- updated variable naming in YouTube component for better distinction
- removed unused jwplayer-comp component -- https://github.com/internetarchive/iaux/issues/305


iisa:
React Dev Tools are not showing their names in the console without it.

PR #287, opened 2019-09-10
Scrubber Bar

jbuckner:
**Description**

> Adds a media Scrubber Bar as part of the Radio Player project

**Technical**

> The scrubber bar is built using `lit-element`. The package was set up using the [`open-wc`](https://open-wc.org) recommended testing and development frameworks.

**Testing**

> The reviewer can check out the `scrubber-bar` branch of the repository, run `yarn install`, then `yarn storybook` to see a fully interactive, customizable scrubber bar.

**Evidence**

scrubber


packages/scrubber-bar/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
```
iisa: This presents a new complexity - how do we trickle down the .eslint config to all child packages from top-level of repo? https://github.com/internetarchive/iaux/blob/master/.eslintrc.json

packages/scrubber-bar/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
iisa: Suggestion: we use AGPL-3 license

packages/scrubber-bar/README.md
```diff
@@ -0,0 +1,67 @@
+# \<scrubber-bar>
```
iisa: love this! 👏  great install steps

packages/scrubber-bar/.gitignore
```diff
@@ -0,0 +1,22 @@
+## editors
```
iisa: Suggestion: we may want to also add these up to the main .gitignore

packages/scrubber-bar/karma.bs.config.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
iisa: Question: can we use Jest as our test runner?

packages/scrubber-bar/package.json
```diff
@@ -0,0 +1,76 @@
+{
+  "name": "scrubber-bar",
+  "version": "0.0.0",
+  "description": "Webcomponent scrubber-bar following open-wc recommendations",
+  "author": "scrubber-bar",
+  "license": "MIT",
```
iisa: Suggestion: change to AGPL-3

packages/scrubber-bar/src/scrubber-bar.ts
```diff
@@ -0,0 +1,185 @@
...
+  html,
+  css,
+  customElement,
+  property,
+  PropertyValues,
+  TemplateResult,
+  CSSResult,
+} from 'lit-element';
+
+@customElement('scrubber-bar')
```
iisa: Suggestion: add `;` per house style

packages/scrubber-bar/src/scrubber-bar.ts
```diff
@@ -0,0 +1,185 @@
...
+  userInteracting = false;
+
+  render(): TemplateResult {
+    return html`
+      <div class="container">
+        <input
+          id="slider"
+          type="range"
+          min="0"
+          max="100"
```
iisa: Question: how do we change the `min`, `max` values, if we want to use another range like 1-5?

packages/scrubber-bar/src/scrubber-bar.ts
```diff
@@ -0,0 +1,185 @@
...
+          step="0.1"
+          value=${this.percentComplete}
+          @mousedown=${this.interactionStarted}
+          @mouseup=${this.interactionEnded}
+          @touchstart=${this.interactionStarted}
+          @touchend=${this.interactionEnded}
+          @input=${this.handleSlide}
+          @change=${this.handleSlide} />
+        <div id="webkit-range-input-style"></style>
```
iisa: Question: does this `<div>` need a closing tag?

packages/scrubber-bar/src/scrubber-bar.ts
```diff
@@ -0,0 +1,185 @@
...
+        color: transparent;
+        ${commonTrackDefinitions}
+      }
+
+      input[type='range']::-ms-fill-lower {
+        background-color: ${trackFillColor};
+        ${trackSizeDefinitions};
+      }
+
+      input[type='range']::-ms-tooltip {
```
iisa: 👏 i appreciate the the thoroughness with the cross-browser pseudo selectors 👏

packages/scrubber-bar/src/scrubber-bar.ts
jbuckner: So right now it's based off of a percentage, but I guess I could just pass around the raw value.

iisa: or make that the default if no values are passed

packages/scrubber-bar/.eslintrc
jbuckner: I was able to extend the base iaux `.eslintrc.json` file with only minor fixes to the style so now this one adheres to the top-level one.

packages/scrubber-bar/.editorconfig
```diff
@@ -0,0 +1,29 @@
+# EditorConfig helps developers define and maintain consistent
```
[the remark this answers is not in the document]
jbuckner: This was added by the `open-wc` template to configure the editor with things like indent-spacing and such if the editor supports it. What are your thoughts on it?

packages/scrubber-bar/karma.bs.config.js
jbuckner: So my first go at test configuration used jest, but I kept running into issues with it. I was having trouble importing ES Modules for testing and also with testing web components. The open-wc group suggests using karma and their template worked out of the box so I left it.

I'm happy to switch to jest if we can figure out how to make it test properly, I just spent a lot of time already :(

packages/scrubber-bar/src/scrubber-bar.ts
jbuckner: Interesting.. so I added a semicolon and you can't actually use decorators with a semicolon. I got a `Parsing error: Declaration expected` when I added it.

packages/scrubber-bar/.editorconfig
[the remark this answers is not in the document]
jbuckner: I removed this file

packages/scrubber-bar/src/scrubber-bar.ts
jbuckner: I re-jiggered this altogether and made it a value since that's the lowest common denominator. The value is a percentage if you set the range 0-100 so it works out good :)

packages/scrubber-bar/src/scrubber-bar.ts
jbuckner: Good catch!

packages/scrubber-bar/src/scrubber-bar.ts
```diff
@@ -0,0 +1,207 @@
...
+  private get rangeSlider(): HTMLInputElement | null {
+    return this.shadowRoot && (this.shadowRoot.getElementById('slider') as HTMLInputElement);
+  }
+
+  private get webkitStyle(): HTMLElement | null {
+    return this.shadowRoot && this.shadowRoot.getElementById('webkit-range-input-style');
+  }
+
+  private updateSliderProgress(): void {
+    if (this.webkitStyle) {
```
shaneriley: **Suggestion**: prefer early return

packages/scrubber-bar/src/scrubber-bar.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  @property({ type: Number }) step = 0.1;
+
+  get percentage(): number {
+    const delta: number = this.max - this.min;
+    const minOffset: number = this._value - this.min;
+    return (minOffset / delta) * 100;
+  }
+
+  private _userInteracting = false;
```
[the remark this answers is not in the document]
jbuckner: It looks like there's a [long-standing conversation](https://github.com/microsoft/TypeScript/issues/9950) going on about that. The mindset at the moment seems to be since only Chrome and Opera support it right now, they weren't considering it its final form, even though they acknowledge, it'll likely be similar to that. 

So right now, no you have to use `private` syntax.

On a side note.. the hash symbol seems like such bad syntax to mark private. I'm curious how they landed on that. 😦

PR #288, opened 2019-09-16
Audio player files parser update

iisa:
**Description**
Music Player files parser updates:
- use /metadata schema to do transformations. This will get us closer to allow @mitra42 and DWEB client to user player
- remove collection-based switch to toggle between the 2 parsers we have.

https://webarchive.jira.com/browse/WEBDEV-2638

**Technical**

Parser tests run through `flatten-album-data.test.js` at this time.
We will continue to have 2 different parser files so we can separate logic into something we can understand.

**Testing**

run storybook > go to theatre stories > check this identifier: `national_lampoon_thats_not_funny_thats_sick` > player should show correctly with track listing

### DIFF
--- a/packages/ia-components/sandbox/theatres/with-youtube-spotify/utils/archive-derived-album-parser.js
+++ b/packages/ia-components/sandbox/theatres/with-youtube-spotify/utils/archive-derived-album-parser.js
@@ -30,7 +30,6 @@ const archiveDerivedAlbumParser = ({ fileDirectoryPrefix, files, itemIdentifier 
   const itemPhotoCandidates = {};
   const trackFilesHaveYoutubeSpotify = [];
 
-  debugger;
   const tracks = files.reduce((neededItems = [], currentFile) => {
     const { source, original = '', name: currentFileName } = currentFile;
     const isPrivate = currentFile.private || null;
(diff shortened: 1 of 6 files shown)


packages/ia-components/sandbox/theatres/with-youtube-spotify/utils/archive-derived-album-parser.js
```diff
@@ -25,19 +25,20 @@ import { isValidAudioFile, isValidImageFile } from './utils';
  * @returns { object } : tracks, itemPhotoCandidates, trackFilesHaveYoutubeSpotify
  *
  */
-const archiveLPAlbumParser = ({ fileDirectoryPrefix, files, itemIdentifier }) => {
+const archiveDerivedAlbumParser = ({ fileDirectoryPrefix, files, itemIdentifier }) => {
   let itemPhoto = '';
   const itemPhotoCandidates = {};
   const trackFilesHaveYoutubeSpotify = [];
 
+  debugger;
```
shaneriley: Got a debugger left in here.

packages/ia-components/sandbox/track-lists/track-list.jsx
```diff
@@ -23,7 +23,7 @@ const parseTrackTitle = ({
   isAlbum = false
 }) => {
   if (isAlbum) { return 'Full album'; }
-  const albumCreatorString = albumCreator.join('; ');
+  const albumCreatorString = albumCreator.join ? albumCreator.join('; ') : albumCreator;
```
shaneriley: **Suggestion**: Instead of relying on the only object type to have a `join` method be an array, test it with `Array.isArray`

iisa: thanks for the catch, forgot to switch it out to that

PR #292, opened 2019-09-20
Moves radio element out of label to prevent double event firing

shaneriley:
**Description**

Fixes https://webarchive.jira.com/browse/WEBDEV-2574

**Technical**

The click event on items in the horizontal radio group fired twice because the radio element was inside of the label element. The change event simulates a click event on parent elements. Moving it just before the label and changing the CSS selector chain for checked state prevent the second click event firing.

**Testing**

See Jira ticket above

**Evidence**

single_analytics_tracker


bfalling:
LGTM2!

Issue #260, opened 2019-08-01
Compare IAUX BookReader wrapper with Dweb BookReader wrapper


iisa:
**Task Details**
- DWeb BookReader wrapper makes a call to `BookReaderJSIA` endpoint for manifest while the IAUX BookReader takes the manifest as a prop.

What can we do to consolidate to 1 React wrapped BookReader?

> What needs to be done and why?

Just an overall audit, which may lead to more questions before defining spec & implementation.

iisa:
@mitra42 this is the ticket to investigate how we can consolidate BookReader components.

mitra42:
I'm not sure if you are WFH or PTO, so I'll update this thread as I learn more …

Just taking a look …, I'm seeing several things that you could be referring to as the IAUX wrapper. 

`sandbox/threatres/bookreader-theatre` which calls ia-js-client and passes the data to a BookReaderJSIA component.  I'm presuming this isn't the wrapper you mean since AFAIK nothing in ia-js-client really works (I'm not using it for anything, and don't believe @jbuckner is). 

Then we have … 

`sandbox/bookreader-component/bookreader-wrapper-main`  goes straight to new Book-reader
`sandbox/bookreader-component/bookreader-jsia.js` passes manifest (I presume) to BookReaderJSIAinit

Then I have `DwebBookReaderWrapper` which fetches data and calls BookReaderJSIAinit

Note that `bookreader/BookReaderJSIAinit` calls new BookReader after adding lending info etc
 
I’m unsure which of these you’d like to merge DwebBookReaderWrapper with - both of the IAUX alternatives seem incomplete - specifically `bookreader-jsia.js` doesn't do any setup of its options, while `bookreader-theatre` doesnt handle the lending info. 

I don't see a problem integrating with either of them, I'd retain DwebBookReaderWrapper, but just do the fetch, and then wrap whichever of your components we are using. 

Except  ... that if integrated with `bookreader-theatre` then the functionality of JSIAinit has to be incorporated, and if integrated with `bookreader-jsia` then the options need to be setup before calling it ... 

@iisa Can you clarify which you are using and whether I'm right about what is missing from each.

mitra42:
@jbuckner  - do you know which of these components @iisa is using?

mitra42:
@iisa -  could you take a look at the above ...

iisa:
sorry for the delay, `sandbox/bookreader-component/bookreader-wrapper-main` is the only component in play right now.

[...]

PR #296, opened 2019-09-25
Bookreader Wrapper integrate JSIA and removal of obsolete files

mitra42:
**Description**
Adds JSIA support to bookreader-wrapper without changing functionality for non-JSIA

**Technical**
Isa wanted there to be one common Bookreader wrapper - great idea, but Dweb uses JSIA to
get IA specific info, this change adds a jsia property to bookreader-wrapper which if present is the metadata as returned by the jsia call, it then uses a different initializer to create the same UI.

There was some talk about having a bookreader-wrapper-main and bookreader-wrapper-jsia but 
the differences between them are too small for that to make sense, and its easily parameterized 

There are a couple of open questions in the code, mostly for Isa, have been unable to reach her, so just leaving the comments in there, the code works without them being answered, and at least they'll flag the issue in case anyone else maintains it in the future. 
 
**Testing**

**Evidence**
This code is live in dweb-mirror (has been for a couple of weeks)


packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
```diff
@@ -8,9 +8,28 @@ import PropTypes from 'prop-types';
...
+ *   options={ options to override default bookreader options}
+ *   />
+ *
+ * Note will almost certainly need a AJS.theatresize() in caller after this.
+ *
+ * Questions for ISA
+ * - why are defaultStartLeaf and titleLeaf not in defaultOptions ? I also moved them to start of list so can be overridden
+ * - I reinstated IABookReaderMessageWrapper (just for the JSIA case) which I'm guessing is used by lending tools to put their message
+ * - Any reason options.onePage.autofit is defaulting to 'height' not 'auto' (which is what I've seen before
+ * - What is going on with the getPgeURI definition, a comment would be good
```
iisa: @mitra42 great questions, thanks for making this a PR to chat about.

One of the issues we have right now, is that the default params are actually custom for the liner notes in the audio items. Are you able to successfully trump them?

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
```diff
@@ -35,6 +54,7 @@ export default class BookReaderWrapper extends Component {
       searchInsideUrl: '/fulltext/inside.php',
       initialSearchTerm: null,
       imagesBaseURL: '/bookreader/BookReader/images/',
+      //TODO-ISA what is going on here, a comment would be useful ? (same on bookreader-wrapper-main and bookreader-wrapper-jsia)
```
iisa: Yes, definitely handy.  Will get that in to clarify, or abstract as passable options.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
```diff
@@ -43,19 +63,28 @@ export default class BookReaderWrapper extends Component {
...
+      ...defaultOptions,
+      ...options,
     };
-    const br = new BookReader(fullOptions);
-    window.br = br;
-    br.init();
+    // There are two ways to initialize BookReader, eithr through JSIA which includes IA specific lending info
+    // or the simpler initialization without IA.
+    if (this.props.jsia) {
+      BookReaderJSIAinit(this.props.jsia, fullOptions); // Creates window.br
```
iisa: is this a reference to the main site's `BookReaderJSIA.js`?

iisa:
@mitra42 the toggles are fine, have you ran this locally in the IAUX context to ensure that it is compatible with its current use as an audio item's liner notes display?

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
mitra42: Sorry, I don't understand that question, what I meant is that you've got a mechanism to setup defaultOptions and then to override them with "options" set in the properties of the component. 
a) Why pull defaultStartLeaf and titleLeaf out of that mechanism rather than having the lines
```
defaultStartLeaf: 0
titleLeaf: 0
``` 
be inside the definition of defaultOptions.   I didn't change this,. 

And b: if you must have them seperate then why put them first in this structure:
```
      ...defaultOptions,
      ...options,
      defaultStartLeaf: 0,
      titleLeaf: 0,
```
which means that they can't be overridden by options, rather than last i.e .
```
    const fullOptions = {
      defaultStartLeaf: 0,
      titleLeaf: 0,
      ...defaultOptions,
      ...options,
    };
```
So that if specified in the component's properties e.g. <BookReaderWrapper titleLeaf:10> it will override your default.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
mitra42: Yes - this is the principle point of the PR, allow (optionally) passing in as props, the jsia returned by the main site's API and then to use the appropriate initializer in BookReaderJSIA.js  ..

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
iisa: @mitra42 the current options listed is specifically for the liner notes display.

> So that if specified in the component's properties e.g. it will override your default.

that's the point.  It's specific to the liner notes use case.

I have created a ticket to parse it out as importable liner notes options for draw: https://github.com/internetarchive/iaux/issues/300

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
iisa: this will get taken care of with this ticket: https://github.com/internetarchive/iaux/issues/300

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
mitra42: OK - maybe I misunderstood that your request in https://github.com/internetarchive/iaux/issues/260 was for a merged wrapper to handle all the cases, in that case I think either
a) defaultStartLeaf and titleLeaf need to be moved to the consumer 
b) OR they can be here as defaults, but in that case they need to be able to be overridden - which they can't if they take precedence like this.   

I must be missing something, because I really can't understand how putting them at lowest precedence (bottom) changes your code at all unless your consumer is passing in some other value for titleLeaf  or defaultStartLeaf that you want this to override.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
mitra42: Happy to, but sorry can you point me at instructions to run Storybook - I've had it shown to me, but don't know how to start it up.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
```diff
@@ -8,9 +8,28 @@ import PropTypes from 'prop-types';
  * This component expects all necessary scripts for BookReader to be pre-loaded onto the page before use.
  * Please see https://github.com/internetarchive/bookreader for more instructions & examples.
  *
+ * If jsia is specified then it adds Internet Archive specific stuff via the BookreaderJSIA function
+ * the script BookreaderJSIA should previously have been loaded.
+ *
+ * TODO - someone could merge all or part of BookreaderJSIA into here and obsolete BookreaderJSIA ideally without changing the interface
+ * TODO - except Isa doesnt want the Archive's lending logic in here.
```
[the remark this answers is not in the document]
mitra42: Up to you - - I like to leave comments in code about things that we know can or should be done in future, as it makes it more likely someone revising it in future will pick that up.  I also thought the reference to you was relevant, so that someone didn't just think it was a good idea when you had (completely reasonable, but non obvious) objections to putting the JSIA code here.  If you prefer, could revise to ...

TODO - The BookreaderJSIA code cant go in here, because it is InternetArchive specific, future development could be a wrapper, for this, that includes BookreaderJSIA

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
iisa: > you want this to override.

yah, it was definitely an override. should we do #300 before hand or merge this in prior?

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
iisa: this should do it: `cd packages/ia-components && yarn run storybook`

if storybook isn't available, just yarn.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
mitra42: Up to you - I don't understand #300 (because I don't know what is happening with liner notes) but sounds (to my ignorance) as something simple.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
mitra42: Thanks - I was looking for an executable and not finding it :-)
Yes - that worked, it comes up on Page 2, is that correct ?

iisa: should look like this, except with Willie Nelson:

image

mitra42: Yes - it does, except it comes up on page 2, and I'm wondering if it is supposed to come up on page 1

iisa: Yes, it should start on page 1.
Is it because these are not set?
```
defaultStartLeaf: 0
titleLeaf: 0
```

mitra42: Its because the caller of BookReaderWrapper is explicitly passing in titleLeaf:1 defaultStartLeaf: 1  ... Thats coming from TheatreAudioPlayer.props.linerNotes.brOptions  
which I think is coming from album-private-with-spotify-youtube.js (i.e. your test data) 

I.e. I think that the Liner Notes go to page 2 (assuming that this is a 0-basd count) with the new BookReaderWrapper, and only go correctly to page 1, when the Component overrides your bad (I think) test data. 

I could be missing something of course .

iisa:
@mitra42 , let's get this over the finish line 🙂  - what else do we need?  i need liner notes to start leafs at 0, regardless of what manifest says + comment updates..

mitra42:
Sure - Comments is waiting on your reply to my question . 

And if you want to force a start at leaf 0, even if options say something else why don't you make your caller pass in defaultStartLeaf=0 titleLeaf=0, you have control between where you read the manifest and where you call BookReaderWrapper. 

Lets have a quick 5 minute call , because I must be missing something if that isn't the Obvious, Easy, and Only Correct way of fixing this, because if you've  hard coded it in the Wrapper there is no way I can make the manifest work correctly in ANY other situation.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
iisa: thanks for resolving, @mitra42

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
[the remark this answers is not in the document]
iisa: updated

PR #298, opened 2019-09-26
Radio Player: Add Transcript View

jbuckner:
**Description**

> Adds an auto-scrolling Transcript View for displaying closed captioning and search results

**Technical**

> This was created using LitElement

**Testing**

> Clone the repo, check out the `transcript-view` branch, `cd iaux/packages/transcript-view`, run `yarn install`, `yarn test` to run tests, `yarn start` to play with the transcript view in the browser

**Evidence**

> transcript-view


packages/transcript-view/dev_tools/search-results-switcher.ts
```diff
@@ -0,0 +1,71 @@
...
+      }
+
+      a {
+        cursor: pointer;
+        -webkit-touch-callout: none; /* iOS Safari */
+        -webkit-user-select: none; /* Safari */
+         -khtml-user-select: none; /* Konqueror HTML */
+           -moz-user-select: none; /* Firefox */
+            -ms-user-select: none; /* Internet Explorer/Edge */
+                user-select: none; /* Non-prefixed version, currently
```
iisa: question: just curious about alignment?

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+  @property({ type: Boolean }) timerRunning = false;
+
+  private timer = -1;
+
+  private timerDelay = 1000;
+
+  render(): TemplateResult {
+    return html`
+      <div class="container">
+        <div>
```
iisa: Question: are the `<div>`s for display purposes? if not, do we need?

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+      <label>
+        <input type="checkbox" .checked=${this.autoScroll} @change=${this.changeAutoScroll} />
+        Auto Scroll
+      </label>
+    `;
+  }
+
+  get currentTimeTemplate(): TemplateResult {
+    return html`
+      <label>Current Time:</label>
```
iisa: blocker: this label is not associated to its input.

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+      <button @click=${this.setNormalTimerDelay}>1x</button>
+      <button @click=${this.setFastTimerDelay}>2x</button>
+      <button @click=${this.setFasterTimerDelay}>3x</button>
+      <button @click=${this.stopTimer}>Stop Timer</button>
+    `;
+  }
+
+  get transcriptHeightTemplate(): TemplateResult {
+    return html`
+      <label>Transcript Height:</label>
```
iisa: same with this label

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+        @change=${this.handleTranscriptSizeSlide}
+      />
+      ${this.transcriptHeight}px
+    `;
+  }
+
+  get showContextBoxesTemplate(): TemplateResult {
+    return html`
+      <label>
+        <input
```
iisa: 👍

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+          @change=${this.changeShowContextZones}
+        />
+        Show Context Zones
+      </label>
+    `;
+  }
+
+  get topContextHeightTemplate(): TemplateResult {
+    return html`
+      <label>Top Context Size:</label>
```
iisa: label association here as well.

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+    `;
+  }
+
+  private searchResultIndexChanged(e: CustomEvent): void {
+    const event = new CustomEvent('searchResultIndexChanged', {
+      detail: { searchResultIndex: e.detail.searchResultIndex },
+      bubbles: true,
+      composed: true,
+    });
+    this.dispatchEvent(event);
```
iisa: Love a custom event.  Is `this.dispatchEvent` part of the Lit Element base?

packages/transcript-view/src/duration-formatter.ts
```diff
@@ -0,0 +1,27 @@
+import { LitElement, html, customElement, property, TemplateResult } from 'lit-element';
+
+@customElement('duration-formatter')
+export default class DurationFormatter extends LitElement {
```
iisa: question: is this something that can be reused?

packages/transcript-view/src/transcript-view.ts
```diff
@@ -0,0 +1,383 @@
...
+      .time-display {
+        position: absolute;
+        top: 0;
+      }
+
+      .scroll-container {
+        display: flex;
+        overflow-y: auto;
+        font-size: 16px;
+        line-height: 24px;
```
iisa: we have been using `rem` in PB. perhaps we can revisit this as a collective

packages/transcript-view/src/transcript-view.ts
```diff
@@ -0,0 +1,383 @@
...
+    if (!scrollView) {
+      return;
+    }
+
+    const start = scrollView.scrollTop;
+    const change = offset - start;
+    const startTime = performance.now();
+    let now;
+    let elapsed;
+    let t;
```
iisa: what's `t`?

packages/transcript-view/test/promised-sleep.js
```diff
@@ -0,0 +1,3 @@
+export default function promisedSleep(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
```
iisa: this is cool

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+    } else {
+      this.startTimer();
+    }
+  }
+
+  private startTimer(): void {
+    this.stopTimer();
+    this.timerRunning = true;
+    this.timer = window.setInterval(() => {
+      if (this.currentTime >= 445) {
```
shaneriley: What does the value `445` represent?

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+  emitCurrentTimeChangedEvent(): void {
+    const event = new CustomEvent('currentTimeChanged', {
+      detail: { currentTime: this.currentTime },
+      bubbles: true,
+      composed: true,
+    });
+    this.dispatchEvent(event);
+  }
+
+  handleTopContextSlide(e: Event): void {
```
shaneriley: **Suggestion**: Refactor this and `handleBottomContextSlide` to call a parent method that accepts a second argument `property`, allowing you to DRY the code.

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+    this.bottomContextHeight = height;
+    const event = new CustomEvent('bottomContextHeightChanged', {
+      detail: { height },
+      bubbles: true,
+      composed: true,
+    });
+    this.dispatchEvent(event);
+  }
+
+  handleTranscriptSizeSlide(e: Event): void {
```
shaneriley: Looks like this method can be added to the suggestion above.

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+    this.transcriptHeight = height;
+    const event = new CustomEvent('transcriptHeightChanged', {
+      detail: { height },
+      bubbles: true,
+      composed: true,
+    });
+    this.dispatchEvent(event);
+  }
+
+  showHideBottomContext(e: Event): void {
```
shaneriley: **Suggestion**: similar DRY request as above, with the methods `showHideBottomContext` and `showHidetopContext` and the `CustomEvent` name param.

packages/transcript-view/package.json
```diff
@@ -0,0 +1,75 @@
+{
+  "name": "@internetarchive/transcript-view",
+  "version": "0.0.1",
+  "description": "A Scrubber Bar WebComponent for use in scrubbing through media",
```
shaneriley: **Suggestion**: Update description to match the component.

packages/transcript-view/src/transcript-view.ts
```diff
@@ -0,0 +1,383 @@
...
+        ${this.showContextZones ? this.contextZoneDevTemplates : ''}
+
+        <div class="scroll-container" id="scroll-container" @wheel=${this.didScroll}>
+          <div class="col time">
+            ${this.timeDisplayTemplate}
+          </div>
+
+          <div class="col">
+            ${this.autoScrollButtonTemplate}
+            ${(this.entries ? this.entries : []).map((entry: TranscriptEntryConfig) =>
```
shaneriley: **Suggestion**: drop the ternary and use a short circuit operation.

```javascript
(this.entries || []).map
```

packages/transcript-view/src/transcript-view.ts
```diff
@@ -0,0 +1,383 @@
...
+      </div>
+    `;
+  }
+
+  private get autoScrollButtonTemplate(): TemplateResult {
+    return html`
+      <button
+        @click=${this.enableAutoScroll}
+        class="auto-scroll-button"
+        style="display: ${this.autoScroll ? 'none' : 'inline-block'}"
```
shaneriley: **Suggestion**: replace the style property toggle with a class toggle. This will make it easier to change which property(ies) are toggled based on boolean state and make it easier to test in unit test libraries that do not fully render the DOM.

packages/transcript-view/src/transcript-view.ts
```diff
@@ -0,0 +1,383 @@
...
+        position: relative;
+      }
+
+      .auto-scroll-button {
+        position: absolute;
+        left: 0;
+        right: 0;
+        margin: auto;
+        width: 8rem;
+        bottom: 1rem;
```
shaneriley: **Suggestion**: move `bottom` prop definition to be grouped with the other positioning props.

packages/transcript-view/src/transcript-view.ts
```diff
@@ -0,0 +1,383 @@
...
+        background-color: ${autoScrollButtonBackgroundColorCss};
+      }
+
+      .context-overlay {
+        position: absolute;
+        left: 0;
+        width: 100%;
+        height: 0;
+        z-index: -1;
+        display: block;
```
shaneriley: You can remove the redundant `display: block` property since setting `position` to `absolute` changes the element's display to `block` by default.

packages/transcript-view/src/transcript-view.ts
shaneriley: That may be difficult since the idea is to get the same default display from this component no matter where you render it. If we're working with a base value of `20px`, for example, setting this to `font-size: 1.6rem;` would give us a font-size of 32 rather than 16.

packages/transcript-view/src/transcript-view.ts
```diff
@@ -0,0 +1,383 @@
...
+      bubbles: true,
+      composed: true,
+    });
+    this.dispatchEvent(event);
+    this.selectedSearchResultIndex = entry.searchMatchIndex;
+    this.autoScroll = false;
+  }
+
+  private handleCurrentTimeChange(): void {
+    const entries = this.entries ? this.entries : [];
```
shaneriley: Same short circuit operation here as above.

packages/transcript-view/src/transcript-view.ts
```diff
@@ -0,0 +1,383 @@
...
+      (entry: TranscriptEntryConfig) =>
+        this.currentTime >= entry.start && this.currentTime <= entry.end,
+    );
+
+    if (!activeEntry) {
+      this.currentEntry = undefined;
+      return;
+    }
+
+    if (this.currentEntry && this.currentEntry.id === activeEntry.id) {
```
shaneriley: I'm a bit confused about the usage of this conditional. if we have matching IDs between `this.currentEntry` and `activeEntry`, then we return, but the only statement left in this block is an assignment of the active entry to the current entry anyway. Can we just remove this entirely?

packages/transcript-view/src/transcript-view.ts
```diff
@@ -0,0 +1,383 @@
...
+
+      now = performance.now();
+      elapsed = (now - startTime) / 1000;
+      t = elapsed / duration;
+
+      scrollView.scrollTop = start + change * easeInOutQuad(t);
+
+      if (t < 1) {
+        window.requestAnimationFrame(animateScroll);
+      } else if (onDone) {
```
shaneriley: Since you have a default value for `onDone`, won't this alway be true when `t >= 1`?

packages/transcript-view/stories/index.stories.js
```diff
@@ -0,0 +1,44 @@
...
+import TranscriptView from '../lib/src/transcript-view';
+import TranscriptEntryConfig from '../lib/src/models/transcript-entry-config';
+import transcript from './transcript';
+
+const convertedTranscript = transcript.map((entry) => {
+  return new TranscriptEntryConfig(
+    entry.id, entry.start, entry.end, entry.text, entry.is_music || false, entry.search_match_index);
+})
+
+console.log(convertedTranscript);
```
shaneriley: **Suggestion**: remove logs or replace with a system like Mitra implemented for debugging in the dweb components.

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
jbuckner: I could get rid of these. This file is just a dev tool (not used for the actual `transcript-vew`, just for testing its functionality) so I didn't spend much time on it 😅

packages/transcript-view/src/duration-formatter.ts
jbuckner: Indeed it can! I was actually considering pulling it out to its own repo, just was a bit tight on time. 😄

packages/transcript-view/src/transcript-view.ts
jbuckner: Oh yeah actually let me make these CSS variables that can be configured from the consumer. I think I set these when Jim and I were trying to calculate the line height, but we don't need to any more.

packages/transcript-view/src/transcript-view.ts
jbuckner: It's a poorly named variable is what it is. 😅 I'll make it more useful.

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  @property({ type: Boolean }) bottomContextVisible = true;
+
+  @property({ type: Boolean }) autoScroll = true;
+
+  @property({ type: Boolean }) showContextZones = true;
+
+  @property({ type: Boolean }) timerRunning = false;
+
+  private timer = -1;
```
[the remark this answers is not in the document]
jbuckner: Oh yeah you can. I'll add it.

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
jbuckner: I'll add it.. these are all dev options so I didn't spend much time on making this file "correct"

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
jbuckner: This is a dev options file. The 445 is the length of the sample transcript I'm using.

packages/transcript-view/dev_tools/search-results-switcher.ts
```diff
@@ -0,0 +1,71 @@
...
+      }
+    `;
+  }
+
+  goToPreviousResult(): void {
+    if (this.currentResultIndex === 0) {
```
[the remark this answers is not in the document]
jbuckner: `currentResultIndex` is reactive so if I increment/decrement first, then update it again, anything that is reacting to its changed state is going to fire twice.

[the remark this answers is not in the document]
shaneriley: Ah, I see.

packages/transcript-view/src/transcript-view.ts
jbuckner: `onDone` is an optional. The syntax in the method signature isn't defining a default value, it's just how you denote the function signature.

ie `onDone?: () => void` is saying "give me a function (optionally) that takes no parameters and doesn't return anything.

You could do `onDone?: (result: string) => void` which is saying "pass me a function (optionally) that takes a single `string` parameter and returns nothing"

TypeScript 3.7 is introducing this feature called "chained optionals" which will allow you to do something like this `onDone?()` which checks for its existence first, then executes if it exists.

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
jbuckner: I tried DRYing this up, but it got kind of weird with the amount of properties I ended up having to pass around. This file is just a dev options file (doesn't get used in the actual transcript view) so I don't think there's much harm in code repetition here.

shaneriley: Wouldn't it be:

```javascript
handleContextSlide(direction, e: Event): void {
    const target = e.target as HTMLFormElement;
    const height = target.value;
    this.bottomContextHeight = height;
    const event = new CustomEvent(`${direction}ContextHeightChanged`, {
      detail: { height },
      bubbles: true,
      composed: true,
    });
    this.dispatchEvent(event);
}
```

and then you can have `handleBottomContextSlide` call this?

jbuckner: You also need to pass around which context to resize (`this.bottomContextHeight` / `this.topContextHeight`) and I couldn't get that to work quite properly. This also opens up the question of "should there be 2 different context height change events or just one". ie:
```
const event = new CustomEvent(`contextHeightChanged`, {
  detail: { height: height, position: `top`/`bottom` },
  bubbles: true,
  composed: true,
});
```

packages/transcript-view/src/transcript-view.ts
jbuckner: We don't want to change the `currentEntry` if it's the same one that we're already on. This method gets called quite often (many times per second) but we don't need it to update until we reach the next entry. I added a comment about this.

packages/transcript-view/src/duration-formatter.ts
jbuckner: I'm going to make this a follow-up task, but I'm all about pulling this out to its own component!

packages/transcript-view/dev_tools/transcript-view-dev-options.ts
jbuckner: I think I'm going to hold off on this refactor until I have more time. This is a very small use case (only for dev options) and I don't think really needs to be elegant or anything. There's some things I need to figure out too (how to pass around which variable that needs incrementing) and also whether to generalize the event. This also means updating the demo app consumer, which is really just a test app and has no impact on the consumable code.

shaneriley: No prob. I don't think I understand the full context anyway. I thought I saw two methods on the same object, and figured they could be consolidated.

Issue #242, opened 2019-07-30
Replace usage of `assert` with `expect` assertion library in unit tests


shaneriley:
In the interest of creating a universal guide for testing in all IA projects, we should replace any usage of the assert library with Jest's built in expect library of matchers.

iisa:
Currently, the only file left with `assert` is `example.test.js`

Acceptance criteria:
- [ ] swap out `assert` in `example.test.js` to `expect`
- [ ] test successfully runs
- [ ] remove `assert` from `package.json`

kimuradev:
@iisa I will work on this ;)

anishagg17:
@iisa i think this issue is closed now

nektro:
@iisa would you be able to close this issue if this is resolved since the PR was merged?

PR #309, opened 2019-10-02
Replace assert to expect function test

kimuradev:
Swap out assert to expect and remove assert dependency.

**Description**

#242 Replaced 'assert' to 'expect' and removed assert depency.

**Technical**

The method assert was removed and replaced to expect from Jest. Also assert dependency was removed because is not used anymore.

**Testing**

Run 'yarn:test' and check 'example-react-component/example.test.js' is 100% coveraged.

**Evidence**

image


nektro:
@iisa would you be able to close this issue if this is resolved since the PR was merged?

PR #322, opened 2019-10-09
Scrubber Bar: Add TypeScript declarations file and update imports

jbuckner:
**Description**

> This adds a declarations file so other TypeScript projects can easily import the Scrubber Bar, as well as cleaning up some imports

**Technical**

> This is mostly some TypeScript updates like the declarations file

**Testing**

> Check out the `scrubber-bar-updates` branch of the repo, `lerna link`, `lerna bootstrap`, `cd packages/scrubber-bar`, `yarn start` and you can play with the scrubber view. `yarn test` to run the tests.

**Evidence**

> No UI updates

### DIFF
--- a/packages/scrubber-bar/test/scrubber-bar.test.js
+++ b/packages/scrubber-bar/test/scrubber-bar.test.js
@@ -1,6 +1,8 @@
 import { html, fixture, expect, oneEvent } from '@open-wc/testing';
 
-import '../scrubber-bar';
+import '../index';
+
+/* eslint-disable no-unused-expressions */
 
 describe('ScrubberBar', () => {
   it('defaults value to 0', async () => {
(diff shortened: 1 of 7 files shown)


packages/scrubber-bar/test/scrubber-bar.test.js
```diff
@@ -93,11 +104,12 @@ describe('ScrubberBar', () => {
     const inputEvent = new Event('input');
 
     // we have to do this in a setTimeout so the event listener below has a chance to listen
-    setTimeout(() => { rangeSlider.dispatchEvent(inputEvent); });
+    setTimeout(() => {
+      rangeSlider.dispatchEvent(inputEvent);
+    });
```
iisa: without a time set, is it just defaulting to `0`?

jbuckner: Yeah with the scrubber bar, it defaults to a range of `0 - 100`, with a value of `0`, stepping by `0.1`.

PR #323, opened 2019-10-09
Waveform Progress: Update some configurations to help with TypeScript support

jbuckner:
**Description**

> Update some TypeScript things like the declarations file for other TS consumers

**Technical**

> Mostly just TS updates.

**Testing**

> Check out the `waveform-progress-updates` branch of the repo, `cd packages/waveform-progress`, `yarn start` and you can play with the Waveform Progress Bar.

**Evidence**

> No real UI updates, just a default color change

### DIFF
--- a/packages/waveform-progress/package.json
+++ b/packages/waveform-progress/package.json
@@ -5,6 +5,8 @@
   "license": "AGPL-3.0-only",
   "main": "index.js",
   "module": "index.js",
+  "types": "index.d.ts",
+  "preserveSymlinks": "true",
   "scripts": {
     "prepare": "npm run build",
     "build": "tsc -p tsconfig.build.json",
--- a/packages/waveform-progress/package.json
+++ b/packages/waveform-progress/package.json
@@ -6,7 +6,6 @@
   "main": "index.js",
   "module": "index.js",
   "types": "index.d.ts",
-  "preserveSymlinks": "true",
   "scripts": {
     "prepare": "npm run build",
     "build": "tsc -p tsconfig.build.json",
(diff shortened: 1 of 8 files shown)


packages/waveform-progress/package.json
```diff
@@ -5,6 +5,8 @@
   "license": "AGPL-3.0-only",
   "main": "index.js",
   "module": "index.js",
+  "types": "index.d.ts",
+  "preserveSymlinks": "true",
```
iisa: what is this toggle for? lerna?

packages/waveform-progress/src/waveform-progress.ts
```diff
@@ -115,7 +115,7 @@ export default class WaveformProgress extends LitElement {
 
   static get styles() {
     const fillColorCss = css`var(--fillColor, #3272b6)`;
-    const zoneOfSilenceColorCss = css`var(--zoneOfSilenceColor, orange)`;
+    const zoneOfSilenceColorCss = css`var(--zoneOfSilenceColor, #f6e652)`;
```
iisa: can we make this a variable?

jbuckner: This is already a variable, I just changed the default value for it to match the one from the mocks. You can override like this:
```
<style>
  waveform-progress {
    --zoneOfSilenceColor: purple;
  }
</style>
```

packages/waveform-progress/package.json
jbuckner: Oh I forgot to remove this, but yeah since you use symlinks in Lerna, there's a flag you have to add when you're using symlinks in your node_modules directory to tell the server to treat the symlinks as regular directories. Otherwise the server gives you an error that the files are outside of the current directory. I don't actually need this so I removed it.

PR #324, opened 2019-10-09
Audio Element: Add Types and Declaration File

jbuckner:
**Description**

> This adds a few types to the audio-element and adds a declarations file so other TypeScript projects can use it as a typed object.

**Technical**

> Mostly just TypeScript-related updates

**Testing**

> Check out the `audio-element-updates` branch of the repo, `cd packages/audio-element`, `yarn start` and you can play with the Audio Element.

**Evidence**

> n/a

### DIFF
--- a/packages/audio-element/index.js
+++ b/packages/audio-element/index.js
@@ -1,3 +1,4 @@
 import AudioElement from './lib/audio-element.js';
+import AudioSource from './lib/models/audio-source.js';
 
-export default AudioElement;
+export { AudioElement, AudioSource };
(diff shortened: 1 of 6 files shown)


packages/audio-element/index.js
```diff
@@ -1,3 +1,4 @@
 import AudioElement from './lib/audio-element.js';
+import AudioSource from './lib/models/audio-source.js';
 
-export default AudioElement;
+export { AudioElement, AudioSource };
```
iisa: question: does eslint bark at this?

jbuckner: Not that I've seen. What warning are you thinking it'd be throwing?

This file just acts as the module entry point so when you do `import { AudioElement } from @internetarchive/audio-element`, it knows where to find the actual class you're looking for.

PR #325, opened 2019-10-09
Playback Controls: Playback Speed, Volume, TypeScript Updates

jbuckner:
**Description**

> Adds playback speed, volume control, and add TypeScript declarations

**Technical**

> This adds a TypeScript declarations file so other TS projects can easily consume it

**Testing**

> Check out the `playback-controls-updates` branch of the repo, `lerna link`, `lerna bootstrap`, `cd packages/playback-controls`, `yarn start` and you can play with the playback controls.

**Evidence**

> 
Screen Shot 2019-10-08 at 5 13 36 PM

### DIFF
--- a/packages/playback-controls/index.d.ts
+++ b/packages/playback-controls/index.d.ts
@@ -1,0 +1,4 @@
+import PlaybackControls from './lib/playback-controls.d';
+import { PlaybackMode } from './lib/playback-mode.d';
+
+export { PlaybackControls, PlaybackMode };
--- a/packages/playback-controls/demo/index.html
+++ b/packages/playback-controls/demo/index.html
@@ -25,7 +25,7 @@
 
   <script type="module">
     import { html, render } from 'lit-html';
-    import '../playback-controls.js';
+    import '../index.js';
 
     render(
       html`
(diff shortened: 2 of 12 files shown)


packages/playback-controls/index.d.ts
```diff
@@ -0,0 +1,4 @@
+import PlaybackControls from './lib/playback-controls.d';
+import { PlaybackMode } from './lib/playback-mode.d';
```
iisa: question: what does the `d` stand for?

jbuckner: TypeScript generates [Declaration Files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) that allow other TypeScript projects to understand what types of objects it is dealing with. Since TypeScript projects get compiled to JavaScript when they are installed, they lose their type information. The `.d.ts` file keeps that type information around so the TS consumers can know what they're dealing with.

jbuckner:
Regarding transpilation in petabox, it will work like this:
1. TypeScript will compile to JS when it is installed from a `yarn install`. There's a `prepare` script in `package.json` that triggers this.
2. Pika will package up up the module and put it in `components/web_modules`
3. Webpack will package that up so we have a transpiled version available when needed

PR #331, opened 2019-10-15
Radio Player

jbuckner:
**Description**

> This is the main Radio Player component that orchestrates all of the other components to make the whole thing work.

**Technical**

> This is written in LitElement and pulls in a bunch of other iaux components.

**Testing**

> Check out the `radio-player-take-two` branch, `cd packages/radio-player`, `yarn install`, `yarn start` to test it in the browser.

**Evidence**

> Screen Shot 2019-10-15 at 10 25 52 AM


packages/radio-player/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+    --trackColor: black;
+    --trackBorder: 1px solid white;
+  }
+</style>
+
+<radio-player
+  .config=${radioConfig}
+  .transcriptConfig=${transcriptConfig}
+  @searchRequested=${this.searchRequested}
```
shaneriley: Can these listener methods be the ones used by default rather than explicitly passing each in?

packages/radio-player/src/search-bar/assets/img/clear-results.ts
```diff
@@ -0,0 +1,7 @@
+import { html } from 'lit-html';
```
shaneriley: Why are these SVG objects converted to lit-html templates rather than just referenced via `<img src` or `background-image`? There don't appear to be any optional arguments to them to allow us to change attributes before using them.

packages/radio-player/src/search-bar/search-bar.ts
```diff
@@ -0,0 +1,248 @@
...
+    const event = new CustomEvent('inputchange', {
+      detail: { value },
+      bubbles: true,
+      composed: true,
+    });
+    this.dispatchEvent(event);
+  }
+
+  private emitEnterKeyPressedEvent(): void {
+    const value = this.searchInput && this.searchInput.value;
```
shaneriley: **Suggestion**: change this line to a ternary that uses either `value` or `''` for the assignment, then on line 94 simply pass the variable as a shorthand property `{ value }`

packages/radio-player/src/search-results-switcher.ts
```diff
@@ -0,0 +1,77 @@
...
+@customElement('search-results-switcher')
+export default class SearchResultsSwitcher extends LitElement {
+  @property({ type: Number }) numberOfResults = 0;
+
+  @property({ type: Number }) currentResultIndex = 0;
+
+  render(): TemplateResult {
+    return html`
+      <div class="container">
+        <a @click=${this.goToPreviousResult}>&lt;</a>
```
shaneriley: Are we changing the design of these with a later commit? I thought it was brought up during the demo that this resembled brackets around a data set rather than controls.

packages/radio-player/src/search-results-switcher.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  updated(changedProperties: PropertyValues): void {
+    if (changedProperties.has('numberOfResults')) {
+      this.currentResultIndex = 0;
+    }
+  }
+
+  goToPreviousResult(): void {
+    if (this.currentResultIndex < 1) {
```
shaneriley: Might be more of a preference, but the way I've always achieved pagination looping is by incrementing/decrementing every call, then checking the value to see if it needs to wrap as one condition rather than two. Something like:
```javascript
this.currentResultIndex -= 1;
if (this.currentResultIndex < 0) {
  this.currentResultIndex = this.numberOfResults - 1;
}
```

packages/radio-player/demo/transcript3.ts
```diff
@@ -0,0 +1,1493 @@
+export default [
```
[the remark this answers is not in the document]
jbuckner: So I need this to get compiled for the demo/dev app and the compiler just takes `.ts` files right now. It basically just copies it to the `lib/` directory and renames it to `.js`

packages/radio-player/README.md
shaneriley: From the look of it, the README is suggesting passing the radio player instance's methods into itself to attach as event handlers. Instead, can we attach them as defaults in the render method and accept optional overrides for each? I haven't used lit-element other than tinkering with it so far, so I'm not sure if this is easily accomplished.

jbuckner: Ohhh I think I see what you're saying. I think this might be where the lines of demarcation get a bit fuzzy between the "radio player" and its host. The radio player itself doesn't handle any searching (or really any data-type events), it just sends the `searchRequested`, `searchCleared`, etc events up to its host and then the host is responsible for executing the search and rebuilding the transcript / search results to pass back down into the radio player. This is what the `radio-player-controller` is doing. It's acting as the bridge between the host and the radio player.

I will add a bit more to this readme, though, to make it more clear what's going on here.

Did I answer your question?

shaneriley: Ah, yeah. I thought those methods were already present on the component, but if they're supplied by whatever controller you're creating to interact with the view layer it makes sense.

jbuckner:
Closing in favor of #343

PR #338, opened 2019-10-29
Scrubber Bar: Add Section Markers

jbuckner:
**Description**

> This adds optional section markers to the scrubber that denote different sections in the playback.

**Technical**

> n/a

**Testing**

> Run `yarn install`, `yarn test` to run the tests. Run `yarn start` to play with it in the browser.

**Evidence**

> Screen Shot 2019-10-29 at 3 48 01 PM

### DIFF
--- a/packages/scrubber-bar/demo/index.html
+++ b/packages/scrubber-bar/demo/index.html
@@ -4,7 +4,15 @@
   <meta charset="utf-8">
   <style>
     body {
-      background: #fafafa;
+      background: black;
+      color: white;
+    }
+
+    scrubber-bar {
+      --trackFillColor: rgba(50, 114, 182, 1);
+      --trackColor: rgba(255, 255, 255, 0.2);
+      --markerHeightExpanded: 25px;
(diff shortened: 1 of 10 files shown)


packages/scrubber-bar/demo/index.html
```diff
@@ -17,10 +25,19 @@
 
     render(
       html`
-        <scrubber-bar></scrubber-bar>
+        <scrubber-bar
+          sectionMarkerPercentages='[0, 10, 11, 25, 30, 50, 75, 100]'
+          value='37'
+          @valuechange=${valueChange}>
+        </scrubber-bar>
```
bfalling: QUESTION: So you're okay with the newline between the opening `<scrubber-bar>` and closing `</scrubber-bar>`. Same question applies down below when the closing tag is put on a different line from the opening tag.

jbuckner: I think this formatting helps to distinguish the parts or the tag, but I'm open to other styles if you have ideas!

PR #341, opened 2019-10-29
Bump Version Dependencies for `ia-prototype-apps`

jbuckner:
**Description**

> Just bump the versions radio dependencies in `ia-prototype-apps`

**Technical**

> n/a

**Testing**

> n/a

**Evidence**

> n/a


packages/ia-prototype-apps/package.json
```diff
@@ -4,11 +4,11 @@
   "license": "AGPL-3.0-only",
   "private": true,
   "dependencies": {
-    "@internetarchive/audio-element": "^0.0.2",
+    "@internetarchive/audio-element": "0.0.1-alpha.422+18a94d7",
```
iisa: question: should we do carrots `^` in prototype apps?

jbuckner: We only have the canary versions on npmjs right now, but yeah ultimately we'd want to use carets `^` once we start pushing non-canary versions

PR #348, opened 2019-11-06
Radio Player: Expandable Search Bar

jbuckner:
**Description**

> This adds a search bar that supports a drop down menu with quick terms.

**Technical**

> This uses LitElement

**Testing**

> Check out this branch:

- cd packages/expandable-search-bar,
- yarn install
- yarn test
- open coverage/index.html
- yarn start

**Evidence**

> 67900524-e53d4080-fb21-11e9-95ea-3771ba92d2d0

Depends on #347


packages/expandable-search-bar/index.js
```diff
@@ -0,0 +1,5 @@
+import ExpandableSearchBar from './lib/expandable-search-bar.js';
```
iisa: this is identical to the `index.d.ts` - i don't think it was identical in the `radio-player` PR - should they be the same like this file?

packages/expandable-search-bar/src/expandable-search-bar.ts
```diff
@@ -0,0 +1,261 @@
...
+  render(): TemplateResult {
+    return html`
+      <div
+        class="
+          container
+          ${this.isOpen ? 'is-open' : ''}
+          ${this.showsDisclosure ? 'shows-disclosure' : ''}"
+      >
+        <div class="search-bar ${this.searchTerm === '' ? '' : 'is-searching'}">
+          <div class="magnifier-container endcap">
```
iisa: can we wrap the `svg` in an image tag to give it proper `alt` text?

packages/expandable-search-bar/test/search-bar.test.js
```diff
@@ -0,0 +1,134 @@
...
+    const quickSearch = el.shadowRoot.querySelector('quick-search');
+    const quickSearchEntryLink = quickSearch.shadowRoot.querySelector('li a');
+    const event = new MouseEvent('click');
+
+    setTimeout(() => { quickSearchEntryLink.dispatchEvent(event); });
+    const response = await oneEvent(el, 'quickSearchSelected');
+
+    expect(response.detail.quickSearchEntry).to.eql(entry1);
+    expect(el.isOpen).to.equal(false);
+  });
```
iisa: 👍 to all of the tests!

packages/expandable-search-bar/test/promised-sleep.js
```diff
@@ -0,0 +1,3 @@
+export default function promisedSleep(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
```
iisa: I didn't see this get imported in. should we remove?
also, what is this function for, primarily?

packages/expandable-search-bar/src/expandable-search-bar.ts
```diff
@@ -0,0 +1,261 @@
...
+import magnifyingGlassIcon from './assets/img/magnifying-glass';
+import disclosureTriangleIcon from './assets/img/disclosure';
+import clearResultsIcon from './assets/img/clear-results';
+
+import QuickSearchEntry from './models/quick-search-entry';
+
+import './quick-search';
+
+@customElement('expandable-search-bar')
+export default class ExpandableSearchBar extends LitElement {
```
iisa: not sure if JSDocs were updated across PRs. I don't see it here...

packages/expandable-search-bar/src/expandable-search-bar.ts
jbuckner: I'm writing out the full `<svg>` tag inline here so I can't wrap that in an `` tag _but_ I learned today that you can add a `<title>` tag to the `<svg>` and it will be used as title text and screen readers will use it. Some more deets on svg accessibility: https://css-tricks.com/accessible-svgs/#article-header-id-1

Screen Shot 2019-11-12 at 11 03 46 AM

packages/expandable-search-bar/src/expandable-search-bar.ts
jbuckner: They're there now.

packages/expandable-search-bar/index.js
jbuckner: It's _nearly_ identical ;) The `.d.ts` file imports the `.d` "declaration" files and this one imports the actual javascript "implementation" files.. ie: 

```
import ExpandableSearchBar from './lib/expandable-search-bar.d';  <-- /.d
vs
import ExpandableSearchBar from './lib/expandable-search-bar.js';  <-- /.js
```

packages/expandable-search-bar/test/promised-sleep.js
jbuckner: Removed. It's used to sleep execution without blocking the thread. I typically use it when I test calling some method on one of elements where the update is going to be asynchronous, but I need to wait for it to update to check the results.

I was trying to find a solution through LitElement, like some event to let me know that the update has happened, but I couldn't find anything that didn't seem super hacky.

packages/expandable-search-bar/src/assets/img/disclosure.ts
```diff
@@ -2,6 +2,7 @@ import { html } from 'lit-html';
 
 const image = html`
   <svg height="9" viewBox="0 0 19 9" width="19" xmlns="http://www.w3.org/2000/svg">
+    <title>Toggle Quick Search</title>
```
iisa: yep, this is perfect. no `img` wrap needed

PR #349, opened 2019-11-06
Radio Player Component Scaffolding

jbuckner:
**Description**

> Add the radio player component npm scaffolding

**Technical**

> This just creates the npm scaffolding for the radio player component

**Testing**

> n/a

**Evidence**

> n/a


packages/radio-player/package.json
```diff
@@ -0,0 +1,91 @@
...
+  "name": "@internetarchive/radio-player",
+  "version": "0.0.0",
+  "description": "A Radio Player LitElement.",
+  "license": "AGPL-3.0-only",
+  "main": "index.js",
+  "module": "index.js",
+  "types": "index.d.ts",
+  "files": [
+    "lib/**/*",
+    "!lib/demo",
```
iisa: question: is the `!` intentional?

packages/radio-player/package.json
```diff
@@ -0,0 +1,91 @@
...
+    "index.d.ts",
+    "index.js"
+  ],
+  "scripts": {
+    "prepare": "npm run build",
```
iisa: Suggestion: Use `yarn` to stay in parity with previous iaux components. We also decided yarn was our main package manager, perhaps we need to revisit

packages/radio-player/package.json
jbuckner: Yeah. This is the list of files that we want to package into the npm module when we upload it to npmjs.io and we want to ignore the demo directory.

packages/radio-player/package.json
jbuckner: Done

PR #343, opened 2019-10-29
Radio Player

jbuckner:
**Description**

> This is the main Radio Player component that orchestrates all of the other components to make the whole thing work. Replaces #331 

**Technical**

> This is written in LitElement and pulls in a bunch of other iaux components.

**Testing**

> Check out the `radio-player-part-deux` branch, `cd packages/radio-player`, `yarn install`, `yarn start` to test it in the browser.

**Evidence**

> Screen Shot 2019-10-29 at 4 21 02 PM

PR #350, opened 2019-11-06
Radio Player Component

jbuckner:
**Description**

> This is the main Radio Player component that orchestrates all of the other components to make the whole thing work. Replaces #343 

**Technical**

> This is written in LitElement and pulls in a bunch of other iaux components

**Testing**

> Check out this branch:

- cd packages/radio-player
- yarn install
- yarn test
- open coverage/index.html
- yarn start

**Evidence**

> 67817103-82886e00-fa69-11e9-8d8c-dc6dc88bc230


packages/radio-player/index.js
```diff
@@ -0,0 +1,16 @@
+import RadioPlayer from './lib/src/radio-player.js';
```
iisa: question: what is the difference between this `index.js` and `index.d.ts`?

packages/radio-player/src/assets/img/left.ts
```diff
@@ -0,0 +1,17 @@
+import { html } from 'lit-html';
+
+const image = html`
```
iisa: question: what is the difference between `left.svg` and `left.ts`?

packages/radio-player/src/assets/img/right.ts
```diff
@@ -0,0 +1,17 @@
+import { html } from 'lit-html';
```
iisa: question: what is the difference between `right.svg` and `right.ts`?

packages/radio-player/src/assets/img/right.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const image = html`
+  <svg height="13" viewBox="0 0 8 13" width="8" xmlns="http://www.w3.org/2000/svg">
```
iisa: question: when the svgs are used, are you wrapping them with an `img` tag to define its usage?

packages/radio-player/src/models/music-zone.ts
```diff
@@ -0,0 +1,10 @@
+export default class MusicZone {
```
iisa: what does this do? would it be possible to add some JSDoc?

packages/radio-player/src/models/radio-player-config.ts
```diff
@@ -0,0 +1,31 @@
+import { AudioSource } from '@internetarchive/audio-element';
+
+export default class RadioPlayerConfig {
```
iisa: would it be possible to prepend the classes with a quick JSDoc on its usage

packages/radio-player/src/radio-player.ts
```diff
@@ -0,0 +1,738 @@
...
+import { QuickSearchEntry } from '@internetarchive/expandable-search-bar';
+import '@internetarchive/expandable-search-bar';
+
+import './search-results-switcher';
+
+import { ZoneOfSilence } from '@internetarchive/waveform-progress';
+import { PlaybackControls, PlaybackMode } from '@internetarchive/playback-controls';
+import SearchResultsSwitcher from './search-results-switcher';
+import MusicZone from './models/music-zone';
+import RadioPlayerConfig from './models/radio-player-config';
```
iisa: 💪 imports!

packages/radio-player/src/radio-player.ts
```diff
@@ -0,0 +1,738 @@
...
+        <div class="waveform-scrubber-container">
+          ${this.waveFormProgressTemplate} ${this.scrubberBarTemplate}
+        </div>
+        ${this.searchSectionTemplate} ${this.transcriptViewTemplate}
+      </section>
+    `;
+  }
+
+  play(): void {
+    /* istanbul ignore else */
```
iisa: question: what is istanbul?

packages/radio-player/src/radio-player.ts
```diff
@@ -0,0 +1,738 @@
...
+    `;
+  }
+
+  private get zonesOfSilence(): ZoneOfSilence[] {
+    /* istanbul ignore else */
+    if (this.duration === 0) {
+      return [];
+    }
+
+    // eslint-disable-next-line max-len
```
iisa: we should update our max-len rule

packages/radio-player/src/radio-player.ts
```diff
@@ -0,0 +1,738 @@
...
+        >
+        </transcript-view>
+      </div>
+    `;
+  }
+
+  private get searchSectionTemplate(): TemplateResult {
+    // The mobile and desktop search sections work similarly, but the mobile version has
+    // a dropdown area on it and the desktop version does not.
+    // This is a case where the functionality is different enough to have two instances
```
iisa: 👍

packages/radio-player/src/radio-player.ts
```diff
@@ -0,0 +1,738 @@
...
+        @searchResultIndexChanged=${this.searchResultIndexChanged}
+      >
+      </search-results-switcher>
+    `;
+  }
+
+  private get noSearchResultsTemplate(): TemplateResult {
+    return html`
+      <div
+        class="no-search-results-message ${this.shouldShowNoSearchResultMessage ? '' : 'hidden'}"
```
iisa: is `hidden` = `display: none`?

packages/radio-player/src/radio-player.ts
```diff
@@ -0,0 +1,738 @@
...
+      </div>
+    `;
+  }
+
+  private get quickSearches(): QuickSearchEntry[] {
+    if (!this.config) {
+      return [];
+    }
+
+    const entries: QuickSearchEntry[] = this.config.quickSearches.map(
```
iisa: q: is it given that `quickSearches` is always within `this.config`?

packages/radio-player/src/radio-player.ts
```diff
@@ -0,0 +1,738 @@
...
+  private emitSearchClearedEvent(): void {
+    const event = new Event('searchCleared');
+    this.dispatchEvent(event);
+  }
+
+  private searchResultIndexChanged(e: CustomEvent): void {
+    if (!this.transcriptView) {
+      return;
+    }
+    this.transcriptView.selectedSearchResultIndex = e.detail.searchResultIndex;
```
iisa: will there always be `searchResultIndex`?

packages/radio-player/src/radio-player.ts
```diff
@@ -0,0 +1,738 @@
...
+  }
+
+  private get transcriptEntries(): TranscriptEntryConfig[] {
+    return this.transcriptConfig ? this.transcriptConfig.entries : [];
+  }
+
+  private get transcriptView(): TranscriptView | null {
+    /* istanbul ignore next */
+    return this.shadowRoot
+      ? (this.shadowRoot.querySelector('transcript-view') as TranscriptView)
```
iisa: I believe that `this.shadowRoot.querySelector` will return `null` if not found, so this (and others) can be shortened to `(this.shadowRoot.querySelector('transcript-view') as TranscriptView) || null`

packages/radio-player/src/radio-player.ts
```diff
@@ -0,0 +1,738 @@
...
+
+  private get searchResultsSwitcher(): SearchResultsSwitcher | null {
+    /* istanbul ignore next */
+    return this.shadowRoot
+      ? (this.shadowRoot.querySelector('search-results-switcher') as SearchResultsSwitcher)
+      : null;
+  }
+
+  private changePlaybackRate(e: CustomEvent): void {
+    this.playbackRate = e.detail.playbackRate;
```
iisa: (maybe it's my JS ptsd,  but object traversal without an error handler makes me hyperventilate)

May be better to use object destructure to ensure that there is a value.
Like: 
```
const { detail } = e;
const { playbackRate = 1 } = detail;
```

packages/radio-player/src/search-results-switcher.ts
```diff
@@ -0,0 +1,81 @@
...
+@customElement('search-results-switcher')
+export default class SearchResultsSwitcher extends LitElement {
+  @property({ type: Number }) numberOfResults = 0;
+
+  @property({ type: Number }) currentResultIndex = 0;
+
+  render(): TemplateResult {
+    return html`
+      <div class="container">
+        <a @click=${this.goToPreviousResult} id="previous-button">${leftImage}</a>
```
iisa: suggestion: use a `button` since it is not a link or page anchor.
is `id` absolutely necessary?
suggestion: wrap `svg` in image tag to define what the image is used for

packages/radio-player/test/radio-player.test.js
```diff
@@ -0,0 +1,630 @@
...
+    const audioSources = [audioMp3];
+
+    const config = new RadioPlayerConfig('foo-title', 'bar-date', '', '', audioSources);
+
+    const el = await fixture(html`
+      <radio-player .config=${config}></radio-player>
+    `);
+
+    // we have to do this in a setTimeout so the event listener below has a chance to listen
+    setTimeout(() => { el.play(); });
```
iisa: 👍

packages/radio-player/test/radio-player.test.js
```diff
@@ -0,0 +1,630 @@
...
+    const searchResultsSwitcher = el.shadowRoot.querySelector('search-results-switcher');
+
+    el.searchTerm = 'foo';
+    el.updateSearchResultSwitcher();
+
+    expect(el.shouldShowNoSearchResultMessage).to.equal(true);
+    expect(el.shouldShowSearchResultSwitcher).to.equal(false);
+  });
+
+});
```
iisa: awesome test suite! 👏

packages/radio-player/src/assets/img/left.ts
iisa: I didn't see the `ts` versions used, perhaps delete them?

packages/radio-player/index.js
jbuckner: The `index.js` file is the entry point for the main `radio-player` module. `index.d.ts` is the entry point for the [typescript declarations](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html), which provide the typescript types for any typescript consumers. 

When you distribute a typescript project, it first gets compiled to javascript so you lose all of the typing. If you're using javascript to consume the module, it doesn't matter since you don't have types, but if you're using typescript to consume the module, it's nice to have the type info and that's what the `.d.ts` provides. It's basically the "header file" for the javascript.

packages/radio-player/src/assets/img/left.ts
jbuckner: Ah actually we use the `left.ts` one because it's kind of a pain to distribute `svg` files with npm. I'll delete the `svg` files.

packages/radio-player/src/radio-player.ts
jbuckner: This is a bit of a typescript oddity that was solved in typescript 3.7 (which was just released a few days ago).

In this case (and you'll see a lot of this going on in the various components) the `shadowRoot` is nullable so you have to first check for the `shadowRoot` before you can call `querySelector` on it.

Typescript 3.7 introduced this awesome feature called optional chaining that let's you do this.. the `?` after `shadowRoot` will just let it return `null` if `shadowRoot` is `null`:

```
return this.shadowRoot?.querySelector('transcript-view') as TranscriptView
```

packages/radio-player/demo/radio-player-controller.ts
```diff
@@ -0,0 +1,242 @@
...
+
+import { AudioSource } from '@internetarchive/audio-element';
+import {
+  TranscriptConfig,
+  TranscriptEntryConfig,
+  TranscriptView,
+} from '@internetarchive/transcript-view';
+
+import transcript from './transcript3.js';
+// prettier-ignore
```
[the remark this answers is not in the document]
jbuckner: Done. Actually I removed it and replaced it with a more targeted ignore.

packages/radio-player/src/assets/img/left.ts
jbuckner: Done.

packages/radio-player/src/assets/img/right.ts
jbuckner: I removed the `svg` files... we were only using the `ts` files

packages/radio-player/src/radio-player.ts
jbuckner: Check em now. I updated all of the spots where we're using CustomEvents to handle failures better and added a bunch o' tests for those cases.

packages/radio-player/src/search-results-switcher.ts
jbuckner: I switched it to use a `button` instead of a link. I need the `id` there to test clicking it. The image now exists inside the button so I hope it's more clear that it's the button's image.

packages/radio-player/test/radio-player.test.js
jbuckner: Thanks!

packages/radio-player/src/assets/img/right.ts
jbuckner: You can just stick them in the DOM from here and they get used wherever you put them. They're like their own self-contained image tags.

packages/radio-player/src/radio-player.ts
jbuckner: Istanbul is the code coverage tool. It's pretty aggressive about its code coverage and if you have an `if` statement without a corresponding `else` clause, it complains that you haven't tested the `else` clause. 

In this case, we only want to call `play()` on the `audioElement` if it exists (it's nullable) but we don't care about the else clause.

This is another case that would get solved by TypeScript 3.7 (which just got released). I could have rewritten this line as: `this.audioElement?.play()`

packages/radio-player/src/radio-player.ts
jbuckner: I'm not totally sure the best way to handle this. There is a framework called `prettier` that auto formats code according to our rules, but it seems like it's having problems keeping this line under our line length limit. It does a pretty good job, but it kept making the next line greater than our length so I ended up just disabling it for this line.

packages/radio-player/src/radio-player.ts
jbuckner: Correct, the radio player config contains quick searches as an array of strings: 
```
config = {
  url: "foo.mp3",
  title: "Bar Channel 3",
  quickSearches: ["foo", "bar", "baz"],
  ...
}
```

This takes those values and converts them into `QuickSearchEntry` objects for use the in the quick search picker.

packages/radio-player/src/radio-player.ts
jbuckner: Yes it is.

Issue #360, opened 2019-12-31
Move all components from ia-components/sandbox => ia-components/components


iisa:
**Task Details**
Currently, we have 2 directories that contain UI components inside of `ia-components` directory: `live` and `sandbox`.

We no longer need the differentiation between the two as all components in sandbox are now live and being used either in archive.org or dweb project.

live vs sandbox was an experiment on staging components into master while they are not fully fleshed out, at this point we see that it is no longer working for us. 

**Task**:
- Move all components from ia-components/sandbox => ia-components/components 
- remove `live` directory
- `components` should be the only directory that houses UI components

**Verification**

- RUN storybook in ia-components and make sure all stories run.
- there is no more `live` directory
- there is only `components` directory

mitra42:
Whats the ETA on this iisa - would like to do a PR with all my stuff to merge so that happens before the refactor ?

iisa:
@mitra42 no hard time frame, just starting to create tickets around repo clean up. we will definitely get your fixes in before this task gets in.

mitra42:
Great - is making ia-components its own repo a separate task for later (I.e. pulling it out of the lerna mess)

RisingGeek:
Is anybody doing this task? If not, @iisa Could you please assign the task to me?

crew-guy:
Hi ! Is this issue resolved ? I would really like to work on it .... 😄

iisa:
hi, thanks so much for being able to lend a hand. this issue was just resolved.

PR #361, opened 2019-12-31
Mitra 361 omnibus

mitra42:
**Description**
A collection of minor bug fixes and code reviews

**Technical**
Should be merged before #360 

**Testing**
Live on dweb.archive.org and in offline archive


packages/ia-components/sandbox/AnchorDetails.jsx
```diff
@@ -27,27 +26,31 @@ const debug = require('debug')('ia-components:AnchorDetails');
...
-    //TODO-STATE this might have the issue of constructor not being re-run and needing componentDidMount catch
-    this.setState({
-      urlProps: ObjectFilter(this.props, (k, v) => AnchorDetails.urlparms.includes(k)),
-      anchorProps: ObjectFilter(this.props, (k, v) => (!AnchorDetails.urlparms.includes(k) && !['children'].includes(k)))
-    });
+    super(props);
+    this.onClick = this.onClick.bind(this);
+    this.state = {};
+  }
+  static getDerivedStateFromProps(props, unusedState) {
```
iisa: question: With the nature of this lifecycle's constant firing, usually changes are only applied when values have changed - this check must be done manually.  I do not see the manual check here, is this function going to hang IF there are many of these components in a certain view?

packages/ia-components/sandbox/AnchorDetails.jsx
```diff
@@ -27,27 +26,31 @@ const debug = require('debug')('ia-components:AnchorDetails');
...
+    };
   }
 
-  clickCallable(ev) {
+  onClick(ev) {
     // Note this is only called in dweb; !Dweb has a director href
     debug('Clicking on link to details: %s', this.props.identifier);
     DwebArchive.Nav.factory(this.props.identifier, { noCache: this.props.reload, wanthistory: !this.props.reload }); // Ignore promise returned
-    return false; // Dont propogate event
+    ev.preventDefault(); // Dont propogate event
```
iisa: this is ok for now, but should we always prevent propagation?

packages/ia-components/sandbox/details/AnchorDownload.jsx
```diff
@@ -78,36 +78,39 @@ function downloadViaAnchor(el, source, options, cb) {
 }
 
 function reachable({ disconnected, source, identifier, filename }) {
+  // Note that multi element arrays arent reachable if disconnected because they need "compress" which we dont currently support on dweb-mirror
   return (!disconnected
     || (source
-      ? (Array.isArray(source) ? ((source.length === 1) && source[0].downloaded) : source.downloaded )
+      ? (Array.isArray(source) ? source.every(s => s.downloaded) : source.downloaded )
```
iisa: is this just a pure refactor or are you looking for an actual value? `every` seems to returns a boolean, and not the value prior to this change.
`https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every`

packages/ia-components/sandbox/details/AnchorDownload.jsx
```diff
@@ -78,36 +78,39 @@ function downloadViaAnchor(el, source, options, cb) {
...
+    // TODO-NAMING check that these urls get routed - if just anchors then have problems on dweb-mirror
     super(props);
     this.onClick = this.onClick.bind(this);
     // TODO-STATE this might have the issue of constructor not being re-run and needing componentDidMount catch
     // this.props passes identifier which is required for Dweb, but typically also passes tabIndex, class, title
-    this.state = {
-      url: new URL((this.props.filename && typeof this.props.filename === 'string') // filename = foo.jpg
-        ? `https://archive.org/download/${this.props.identifier}/${this.props.filename}`
-        : (this.props.source && Array.isArray(this.props.source) && this.props.source.length === 1) // source = [ArchiveFile]
+    const url = new URL((this.props.filename && typeof this.props.filename === 'string') // filename = foo.jpg
```
iisa: in this file (and others), can we break all these nested ternaries down for better readability/scanning?

packages/ia-components/sandbox/details/AnchorSearch.jsx
```diff
@@ -63,21 +63,24 @@ class AnchorSearch extends IAReactComponent {
...
   constructor(props) {
     super(props); // { query || field & value, sort, reload }
-    //TODO-STATE this might have the issue of constructor not being re-run and needing componentDidMount catch
-    this.setState({
-      query: this.props.query || ObjectFromEntries([[this.props.field, this.props.value]]),
-      urlProps: ObjectFilter(this.props, (k, unusedV) => AnchorSearch.urlparms.includes(k)),  //sort, reload, query
-      anchorProps: ObjectFilter(this.props, (k, unusedV) => (!AnchorSearch.urlparms.includes(k) && !['children'].includes(k)))
-    });
+    this.onClick = this.onClick.bind(this);
+    this.state = {};
```
iisa: what's your constructor preference, setting state before or after the event bindings?

packages/ia-components/sandbox/details/AnchorSearch.jsx
```diff
@@ -63,21 +63,24 @@ class AnchorSearch extends IAReactComponent {
...
+    };
+  }
+  onClick(ev) {
     // Note this is only called in dweb; !Dweb has a director href
     debug('Clicking on link to search: %s', this.state.query);
     DwebArchive.Nav.navSearch(
       this.state.query,
       { noCache: this.props.reload, wanthistory: !this.props.reload, sort: this.props.sort });
-    return false; // Dont propagate event
+    ev.preventDefault(); // Dont propogate event
```
iisa: commenting on this as i am not sure preventing events to flow freely is the right way to go.

packages/ia-components/sandbox/details/CrawlConfig.js
```diff
@@ -44,17 +42,20 @@ const debug = require('debug')('dweb-archive:CrawlConfig');
...
-      level: props.level,
-      clickable:  !CrawlConfig.unclickable.includes(this.props.identifier)
-    })
+    this.onClick = this.onClick.bind(this);
     CrawlConfig.instance = this; // Allow finding it
+    this.state = {};
   }
 
+  static getDerivedStateFromProps(props, unusedState) {
+    return {
```
iisa: adding another comment here as I do not see any checks for valid value changes

packages/ia-components/sandbox/details/CrawlConfig.js
```diff
@@ -113,6 +114,7 @@ export default class CrawlConfig extends IAReactComponent {
         }
       });
     }
+    ev.preventDefault(); // Dont propogate event
```
iisa: adding another comment around prevent default - do we need to do this?

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
```diff
@@ -22,37 +25,40 @@ import { I18nSpan, I18nStr, I18nIcon } from "../languages/Languages";
...
+    return this.props.files
+      .filter(af => ( !this.props.disconnected || af.downloaded))
+      .reduce((res, af) => {
+        const format = af.metadata.format;
+        if (downloadableFormat(format)) {
+          if (!res[format]) { res[format] = []; }
+          res[format].push(af);
+        }
+        return res;
+      }, {});
```
iisa: a few items to note - 
- I don't see a constructor that explicitly binds this class state to the function yet this function is calling `this.props`.
- What happens when `files` is not an array? Will it bork immediately?

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
```diff
@@ -22,37 +25,40 @@ import { I18nSpan, I18nStr, I18nIcon } from "../languages/Languages";
...
   render() {
     // Build a dictionary of file formats
-    //TODO Add the 'reachable' test in Anchor Download to this filter
-    //TODO See https://github.com/internetarchive/dweb-mirror/issues/246 for missing files issue
+    // TODO Add the 'reachable' test in Anchor Download to this filter
+    // TODO See https://github.com/internetarchive/dweb-mirror/issues/246 for missing files issue
 
     const downloadableFilesDict = this.downloadableFilesDict();
     const filesCount = this.props.files_count;
     const originalFilesCount = this.props.files.filter(f => f.metadata.source === 'original').length + 1; // Adds in Archive BitTorrent
```
iisa: suggestion: destructure `this.props`

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
```diff
@@ -71,26 +77,31 @@ export default class DetailsDownloadOptions extends React.Component {
             >
               {formats('format', k).downloadable}
               {' '}
-              <I18nIcon className="iconochive-download" en="download"/>
+              <I18nIcon className="iconochive-download" en="download" />
             </AnchorDownload>
+            {(typeof DwebArchive === 'undefined' || !DwebArchive.mirror || (k !== 'Epub')) ? null : (
```
iisa: what is this for? can this be abstracted out to a function so we can gain quick insight when we read through/use this file?

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
```diff
@@ -71,26 +77,31 @@ export default class DetailsDownloadOptions extends React.Component {
...
+              <a href={`/epubreader/index.html?bookPath=/download/${this.props.identifier}/${downloadableFilesDict[k][0].metadata.name}`} title="Read online">
```
iisa: suggestion: abstract this file URL to a function. this part is my code smell - `downloadableFilesDict[k][0].metadata.name` - what happens when you can't traverse all the way down to get the value? is the link broken?

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
```diff
@@ -71,26 +77,31 @@ export default class DetailsDownloadOptions extends React.Component {
...
+                <I18nIcon className="iconochive-eye" en="Read online" />
+              </a>
+            )}
           </div>
         ))}
         <div className="show-all">
-          {(this.props.disconnected) ? null :
+          {(this.props.disconnected) ? null : (
```
iisa: this can be more succinct with `!this.props.disconnected && <dom>` - for better readability. or can be abstracted to function

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
```diff
@@ -71,26 +77,31 @@ export default class DetailsDownloadOptions extends React.Component {
...
+                <I18nIcon className="iconochive-download" en="download" />
+                {' '}{filesCount}{' '}{I18nStr('Files')}
               </a>
-              <br/>
+              <br />
               <a className="boxy-ttl hover-badge" href={compressAllURL}>
-                <I18nSpan className="iconochive-download" en="download"/>
-                {originalFilesCount}{' '}{I18nStr("Original")}
+                <I18nSpan className="iconochive-download" en="download" />
+                {originalFilesCount}{' '}{I18nStr('Original')}
```
iisa: this can be cleaned up 
`{originalFilesCount}{' '}{I18nStr('Original')}` => {`${originalFilesCount} ${I18nStr('Original')}`}

along with the instance up top. let's use the the backtick way if we are cobbling together string values from variables

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
```diff
@@ -71,26 +77,31 @@ export default class DetailsDownloadOptions extends React.Component {
...
               </a>
-              <br/>
+              <br />
             </div>
-          }
-          <AnchorDownload className="boxy-ttl" identifier={this.props.identifier} disconnected={this.props.disconnected}><I18nSpan en="SHOW ALL"/></AnchorDownload>
+          )}
+          <AnchorDownload className="boxy-ttl" identifier={this.props.identifier} disconnected={this.props.disconnected}><I18nSpan en="SHOW ALL" /></AnchorDownload>
           <br clear="all" className="clearfix" />
```
iisa: ahhh clearfix.  do we need?

packages/ia-components/sandbox/details/ModalGo.jsx
```diff
@@ -18,7 +18,8 @@ class _ModalGo extends React.Component {
   onClick(ev) {
     // ev.currentTarget is the HTML Element on which the onClick sits
     // noinspection JSUnresolvedFunction
-    return AJS.modal_go(ev.currentTarget, this.props.opts);
+    AJS.modal_go(ev.currentTarget, this.props.opts); // Unclear if return from this is what we want (false to run ev.preventDefault)
+    ev.preventDefault(); // Dont propogate event
```
iisa: pinging again to ask the question - do we need to stop event propagation?

packages/ia-components/sandbox/details/NavWrap.jsx
```diff
@@ -88,10 +88,10 @@ class NavSearchLI extends React.Component {
     this.setState({ value: event.target.value });
   }
 
-  onClick(unusedEvent) {
+  onClick(ev) {
     // this.formElement.submit(); // TODO Doesnt work - not sure why
     this.onSubmit(unusedEvent);
-    return false;
+    ev.preventDefault(); // Dont propogate event
```
iisa: pinging again to ask the question - do we need to stop event propagation?

packages/ia-components/sandbox/details/Search.jsx
```diff
@@ -19,31 +21,32 @@ class SearchSwitcher extends IAReactComponent {
...
+          {/* --TODO-DETAILS this dropdown doesnt reorder, test other UI elements in vicinity as well see https://github.com/internetarchive/dweb-archive/issues/15--*/}
+          <span className="big-label blue-pop">{/* TODO-ISSUE dweb-archive#57 remove relevance on Collections */}
+            {(!this.props.identifier) // Dont show on collections
+              ? <AnchorSearch className="ikind stealth in" data-id="relevance" query={this.props.query}><I18nSpan en="RELEVANCE" /></AnchorSearch>
               : undefined}
-            {(!this.props.identifier) ? // Dont show on collections
-              <div className="iconochive-dot ikind-sep"></div>
+            {(!this.props.identifier) // Dont show on collections
+              ? <div className="iconochive-dot ikind-sep" />
               : undefined}
```
iisa: this can be in the same block. `null` is the appropriate value to send if you don't want DOM to be drawn.

Also, you can use the `!this.props.identifier && <DOM>` shortcut - also for better readability

packages/ia-components/sandbox/details/Search.jsx
```diff
@@ -19,31 +21,32 @@ class SearchSwitcher extends IAReactComponent {
...
           </span>
         </div>
       </div>
     );
   }
 }
-export {SearchSwitcher}
\ No newline at end of file
+export { SearchSwitcher };
+/* Code review Mitra 2019-12-11 excluding HTML comparisom */
```
iisa: typo comparison

packages/ia-components/sandbox/details/Tabby.jsx
```diff
@@ -60,22 +58,22 @@ export default class Tabby extends IAReactComponent {
...
-              : <I18nSpan className="tabby-text" en={this.props.text}/>
+              : <I18nSpan className="tabby-text" en={this.props.text} />
             }
           </a>
         </div>
       </div>
     );
   }
 }
+// Code inspection by Mitra 2019-12-29 exc html compare
```
iisa: linter didn't bark without the new line at the end?

packages/ia-components/sandbox/languages/Languages.jsx
```diff
@@ -28,8 +27,16 @@ const languageConfig = { // Note the flags are dragged out of the Mac Emoji and
...
 
+/**
+ * Fetch a supported language
+ * @param lang
+ * @param cb(err)
+ */
 function getLanguage(lang, cb) {
-  if (!languageConfig[lang]) { cb(new Error('Do not support language: '+lang));
+  if (!languageConfig[lang]) {
+    cb(new Error('Do not support language: ' + lang));
```
iisa: does the new Error break DOM drawing?

packages/ia-components/sandbox/tiles/TileComponent.jsx
```diff
@@ -45,9 +45,9 @@ export default class TileComponent extends React.Component {
       const item = this.props.item;
       const query = props.query || (item && item.query) || (member && member.query);
       const isCollection = (member.mediatype === 'collection');
-      const collection0 = member.collection0() || (item && item.metadata && item.metadata.collection[0]);
+      const collection0 = member.collection0() || (item && item.metadata && item.metadata.collection && item.metadata.collection[0]);
       const by = member.creator || member.creatorSorter || (item && item.metadata.creator); // Should be array
-      const collection = member.collection || (item && item.metadata.collection) || []; // Should be array (note fav-xxx dont have collection)
+      const collection = member.collection || (item && item.metadata && item.metadata.collection) || []; // Should be array (note fav-xxx dont have collection)
```
iisa: let's leverage some destructuring..

packages/ia-components/sandbox/AnchorDetails.jsx
mitra42: I'm not sure I understand you ... or how multiple components in review would cause a hang (which suggests I'm missing something about React). 

My understanding was that it was componentDidUpdate that needs the check, because it will fire itself if it changes state while getDerivedStateFromProps only ever fires once on a change. .https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

packages/ia-components/sandbox/AnchorDetails.jsx
mitra42: In this case yes, because otherwise it will also navigate to the href. Note this only happens for Dweb cases. (onClick is not attached for other cases).

packages/ia-components/sandbox/details/AnchorDownload.jsx
mitra42: I'd hate to do that, nested ternaries can be hard to read if used in a fully-freeform manner, but I believe there is a fundamental pattern here of 
```
test1
? result1
: test2 
? result2
: result-else
```
Which is not easily expressed in Javascript in other ways, certainly not more readably.
Which is why I line them up one above the other (though lint tends to rearrange them).

packages/ia-components/sandbox/details/AnchorSearch.jsx
mitra42: Is there are reason for a preference ? It should be exactly equivalent.

packages/ia-components/sandbox/details/AnchorSearch.jsx
mitra42: Same answer as for AnchorDownload - Dweb only, and otherwise its going to trigger the href. (And its useful to have the href there for things like the display in the status bar and so on)

packages/ia-components/sandbox/AnchorDetails.jsx
mitra42: Note - I'm skipping answer on same question on other places below, will leave as placeholder till resolved here ....

packages/ia-components/sandbox/details/CrawlConfig.js
mitra42: This one could go either way, I thought it was preferential not to pass it up when handled, but happy to remove (in this case, not in the Anchor cases).

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
mitra42: I'm not sure the problem here, its a method on the class called from the render (to make render easier to ready), it doesn't need to be bound because as a class method it always has access to this (its only onClick etc which for some react-bizarre reason don't get passed "this") 

It should barf on files not being an array - that would be a coding error and AFAIK we dont expect code to do extreme checking for errors that would be coding errors (as opposed to data errors).

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
mitra42: Can you explain why destructing props is better than an assignment, I've seen that preference expressed in esLint, and have never seen a good explanation of why 
```
const {filesCount} = this.props;
```
is in any way better than 
```
const filesCount = this.props.filesCount
```
IMHO its neither more efficient, nor more readable nor less prone to error.

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
mitra42: Its the condition for why the nested HTML should be included. 
Sure it could be extracted out (and if it was more than one line I would do) but that would put the test (only used once) in a different place of the code than it use which I think would make it less readable. 
```
function isNotEpubOnMirror(format) { 
  return !DwebArchive && !DwebArchive.mirror || (format !== 'Epub')
}
```
But it still has to be used in that form since I believereact requires "null" if something is to be skipped. 
```
{ isNotEpubOnMirror(k) ? null : (
```

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
mitra42: Good question ... but 
- downloadableFilesDict[k] always exists since you are iterating over the keys of downloadableFilesDict. 
- Its always non-empty, because its an inversion, so "k" only exists for members
- files always have metadata which always have a name. 

Sure - abstracted to 
```
function epubReaderUrl(identifier, file) {
    return `/epubreader/index.html?bookPath=/download/${identifier}/${file.metadata.name}`;
  }
```
and
```
href={epubReaderUrl(this.props.identifier, downloadableFilesDict[k][0])}
```

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
mitra42: I use both, but ...  change made as requested though
*  I'm certain its less efficient when you are just appending constant and variable strings. 
*  I think its less readable i.e.  variable inside backtick-escape inside backtick inside react-escape  instead of variable inside react-escape

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
mitra42: I don't know, wherever possible I copy the html that comes out of the existing petabox, and its there in the petabox code (search for ` <br clear="all" class="clearfix"/>` in https://archive.org/details/commute for example.  I periodically review the HTML my code generates against petabox html, but try not to get ahead of petabox code as then I get bitten by changes to archive.css that (quite reasonably) aren't tested against my code

packages/ia-components/sandbox/details/ModalGo.jsx
mitra42: Yes - there's a href in that anchor  that is used for non-dweb cases and would otherwise be triggered for dweb

packages/ia-components/sandbox/details/NavWrap.jsx
mitra42: In this case no - though as I said, I thought it was good practice to catch clicks you've acted on. change made

packages/ia-components/sandbox/details/Search.jsx
mitra42: Change made to use null because I understand that method,  (note the use of undefined is in old code before I knew about null, only change here was lint-driven formating to drop the "?" down.

mitra42: Can you explain that second construct. I believe `!this.props.identifier && <DOM>` returns `false` and I thought you had to send `null` to avoid drawing DOM. If you are sure that is valid I will make that change as I edit code in future.

packages/ia-components/sandbox/details/Search.jsx
mitra42: fixed

packages/ia-components/sandbox/details/Tabby.jsx
mitra42: Yes - usually does, but this comment was added to note that I'd done the linting :-)

packages/ia-components/sandbox/languages/Languages.jsx
mitra42: getLanguage is called much earlier - during language switching, not during DOM drawing

packages/ia-components/sandbox/tiles/TileComponent.jsx
mitra42: How could that line be written more clearly with deconstruction, its a complex expression assigned to a constant ? Any deconstruction is going to require superflous temporary constants, still require all thsoe tests and decrease readability ... feel free to propose an alternative.

traceypooh: ok, i've really *really* disliked deconstruction since the start.

i've never seen another language doing this, and my brain would *full brakes*, pause, try to even remember how these stupid things work, before continuing code scans...

however!  over time, i have _just_ come around the corner to liking them -- and i think preferring them.

when there are 2+ elements, it _really_ becomes a big win (IMHO).  and now i've found myself wishing it was in other langs i use a lot.

eg:
`const { title, description } = metadata`

i find myself no longer wanting to write `description` twice -- and is starting to feel more like clutter, compared to deconstruction.  

$0.02

mitra42: Yes - I'm not pro, or anti, deconstruction - I use them myself especially for multiple items like your example, and for the results of a method/function.
<rant> What I really really hate, is some wet-behind-the-ears coder at AirBnB defining a set of eslint rules, that arbitrarily disallow some particular and perfectly valid / not-confusing coding pattern in favor of something new and shiny, and that then we are adopting those rules as if they are in some way better - I've come to appreciate elint for catching some errors but it needs pruning of entirely arbitrary preferences of whoever wrote them. </rant>

jbuckner: That's the whole point of a style guide, though. It gives all of the developers a common style to develop on. It gets rid of a lot of these bike-shedding conversations during merge requests because the developers have agreed upon a standard.

A conversation can be had about whether or not to enable different features of it like destructuring, but we should follow the style guide that we've chosen.

mitra42: But those two lines don't come close to the one I've got. it makes absolutely no sense to take the line:
```
const collection = member.collection || (item && item.metadata && item.metadata.collection) || [];
```
And replace it with 6 lines, plus a tonne of commenting to explain what doing. 
```
const { metadata = {} } = item;
const { title, collection } = metadata;
// Have to move it out the way, as we have two fields called "collection" 
const itemcollection = collection; 
const { collection } = member;
// Move this one out the way
const membercollection = collection; 
const collection = membercollection || itemcollection || [];
```
This is *nuts* if we've got style guidelines which enforce writing s***y code like this. I strongly recommend that if we are insisting on following all rules in the style guide, that this particular rule get removed.

packages/ia-components/sandbox/AnchorDetails.jsx
iisa: Hi @mitra42 ,
Pardon me for not being clear.  I hope I can do better here.

I am touching on the other two points mentioned in the doc -
> If you want to re-compute some data only when a prop changes
> If you want to “reset” some state when a prop changes

From what you are returning, it seems like you want to do either one of those two points.  Is that off-base?  If so, do let me know.

But, with that assumption + `getDerivedStateFromProps` event firing with every update (click events, state updates, etc.), it seems that running the filter functions on every return can get costly.  Especially if one stacks multiple components on the page with this same pattern.

What the doc recommends may be a bit overkill for our immediate needs, so I am suggesting to only return new state values when the values change. 
-------
What are your thoughts?

packages/ia-components/sandbox/details/AnchorSearch.jsx
iisa: just pickin' your brain around code organization.

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
iisa: Given that `this.props` will always exist in a React component, traversing 1 level is fine, but what happens when we go further?
Things happen, and I don't want one lil JS exception to stop the whole show. 

I think having code that actively guards against unwanted behavior just part of the good JS habits we want to uphold.

> IMHO its neither more efficient, nor more readable nor less prone to error.
Ya, sure. But we are writing ES6, let's lean into it.

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
iisa: would you be able to put it in a variable as a tldr?

packages/ia-components/sandbox/AnchorDetails.jsx
mitra42: The techniques suggested under the `re-compute` and `reset` lines are way complicated, I've never understood either of them. 

You don't have access to previous props in getDerivedStateFromProps, so I can't compare props to see if they have changed.   And once I've compute the new state (filtering props into two buckets), then I don't believe there is any advantage to not returning that state, and there is more computation to do to compare state). Note the operations done are cheap - a filter of the props into two buckets. 

Am I missing something (note I've suggested a call about this, otherwise I think this back and forth with 16 day gaps is going to take forever)

I believe this is the same for the other places I've used this pattern, but I'll recheck them once we get clarity on this case.

packages/ia-components/sandbox/details/AnchorDownload.jsx
mitra42: Yes - I want a boolean - (the ARCHIVEFILE.downloaded field is a boolean), so both cases return a boolean - old code was only checking the first file (for some strange reason i cant figure out), and it needs to check all of them.

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
mitra42: The new features in ES6 don't eliminate all the existing features of javascript.  Someone at AirBnB decided that they prefered destructuring so decided to impose it on everyone else, when (for single values) there is zero advantage AFAIK.  

I'm not sure what you mean by traversing multiple levels - if you mean for example something like 
```
const foo = this.props.bar.baz
```
Then you need to handle the case when `bar` isn't set in props, but checking that is required with destructuring as well - and in fact I wouldn't even know how to write such a two-level destructure. 

Note this is not in code I've changed in this PR, and I'm certainly not going to go back through all my code and make changes to match that particular preference for zero gains in performance, reliability, or error prone-ness.

packages/ia-components/sandbox/details/DetailsDownloadOptions.jsx
mitra42: Please give me a code snipped of what you are thinking, everything I can think of as alternatives is WORSE (More confusing, more error prone) than doing this test in-line at the place its used.

PR #371, opened 2020-01-24
Update README.md

aashimgarg:
i tried to install yarn dirctly as mentioned in the readme but then i faced an error. by running the given command before the yarn install i was not facing any error.


aashimgarg:
sorry i didnt get that.

aashimgarg:
I found the error

PR #376, opened 2020-01-27
Increases desktop max and widescreen min widths

shaneriley:
**Description**

This increases the width range of the desktop breakpoint to account for both BookReader controls being visible and track lists being rendered nicely when long track names are present. Closes [WEBDEV-3056](https://webarchive.jira.com/browse/WEBDEV-3056)

**Testing**

See Jira ticket


shaneriley:
Do we need a version bump for this?

iisa:
> Do we need a version bump for this?

yes, but will do in in a package of fixes. 

feel free to merge in, otherwise :)

PR #379, opened 2020-01-29
Creates language select component

### LINKED ISSUE #378: Create language select
**Context**

DWeb would like to render a nicely styled component to select a language from a dropdown.

**Proposal & Constraints**

Create a React component that accepts an array of options for the dropdown, a string for the value of the currently selected language, and an optional callback that will be called with the newly selected language value as the first parameter. To reduce the number of assets needed to render this component, the country flag icons and selected icon should be rendered as emoji.

**Additional Resources**

See https://webarchive.jira.com/browse/WEBDEV-2794 for more info on the request and design documents.

**Stakeholders**

@mitra42

shaneriley:
**Description**

Closes #378

**Technical**

Sandbox component added to Storybook. Run Storybook using the README instructions to test implementation.

**Testing**

Interact with rendered component in Storybook. Run `yarn test` within packages/ia-components to run unit tests.

**Evidence**

language_dropdown

### DIFF
--- a/packages/ia-components/sandbox/language-select/language-select.js
+++ b/packages/ia-components/sandbox/language-select/language-select.js
@@ -21,13 +21,13 @@ const flags = {
  *
  * @param object props
  *   @param string name language display name
- *   @param string value language value
+ *   @param string flag flag emoji
  *
  * @returns component stateless
  */
-const LanguageNode = ({ name, value }) => (
+const LanguageNode = ({ name, flag }) => (
   <>
-    <span className='flag'>{flags[value]}</span>
+    <span className='flag'>{flag}</span>
     {name}
   </>
 );
@@ -36,17 +36,17 @@ const LanguageNode = ({ name, value }) => (
  * Renders a selectable language option in the languages dropdown
  *
  * @param object props
- *   @param string name language display name
+ *   @param object language language object
  *   @param string value language value
  *   @param string selectedLanguage selected language value
  *   @param function onClick click handler from LanguageSelect instance
  *
  * @returns component stateless
  */
-const LanguageOption = ({ name, value, selectedLanguage, onClick }) => (
+const LanguageOption = ({ language, value, selectedLanguage, onClick }) => (
   <li className={selectedLanguage === value ? 'selected' : null}>
     <a onClick={onClick} data-language={value}>
-      <LanguageNode name={name} value={value} />
+      <LanguageNode name={language.inLocal} flag={language.flag} />
     </a>
   </li>
 );
@@ -58,9 +58,9 @@ const LanguageOption = ({ name, value, selectedLanguage, onClick }) => (
  * opens a dropdown list of languages that render flag icons and the language
  * name written in its native language.
  *
- * @param array options - array of objects to select from
+ * @param object languages - object map of language values
  * input object example:
- * { value: 'en', label: 'English' }
+ * { 'en': { inEnglish: 'English', inLocal: 'English', flag: '🇬🇧' } }
  * @param function onSelect - event handler
  * @param string selectedLanguage - language that is selected
  *
@@ -68,9 +68,9 @@ const LanguageOption = ({ name, value, selectedLanguage, onClick }) => (
  */
 class LanguageSelect extends React.Component {
   constructor(props) {
-    const { options, selectedLanguage, onSelect } = props;
+    const { languages, selectedLanguage, onSelect } = props;
     super(props);
-    this.options = options;
+    this.languages = languages;
     this.state = {
       selectedLanguage,
       dropdownOpen: false,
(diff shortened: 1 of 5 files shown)


packages/ia-components/sandbox/language-select/language-select.js
```diff
@@ -0,0 +1,140 @@
...
+    super(props);
+    this.options = options;
+    this.state = {
+      selectedLanguage,
+      dropdownOpen: false,
+    };
+    this.onSelect = onSelect;
+
+    ['toggleDropdown', 'selectLanguage'].forEach((method) => {
+      this[method] = this[method].bind(this);
```
jbuckner: What does this binding do here? Does the `<a onClick={this.toggleDropdown}>` in the `render()` method do the same thing?

shaneriley: This is a combination of approaches. The first is to ensure we have instance context within our callbacks. The second is to reduce memory footprint by only creating one bound function in memory rather than one for each option in the dropdown. The `toggleDropdown` method doesn't get attached more than once, but I like to keep the binding in one spot for all DOM methods.

mitra42:
This all looks fine to me in general, I'll have to include it on the page to be sure, but I don't see anything that would be a problem. 

One thing I'd do differently, is to use the structure in https://github.com/internetarchive/dweb-archive/blob/7a0381bb557d1987fe764d91ae14402a4401949b/ia-components/sandbox/languages/Languages.jsx#L15 otherwise I'm going to have to produce an in-memory inversion of the  structures you use from 
`{  en: { inEnglish: 'English', inLocal: 'English', flag: '🇬🇧'  },... }`
  to  ` [ { name: English, value: en },... ]` and `{ en: '🇬🇧' ... }` 

If we cant use my structure, could we at least call that field 'inLocal' rather than 'name' which a) makes the constructor easier and b) is more explicit (name isn't clear as to whether you mean name in English or name in its own language.

packages/ia-components/sandbox/language-select/language-select.js
jbuckner: For the first approach, when the callback is made from the `onClick`, isn't it always in the context of the current object, `this`?

For the binding multiple times, does it create an instance for every one? I figured each option was just emitting the `onClick` event and all passing it to the same listener since you can distinguish the callers from `e.target`.

shaneriley: The context will be the DOM element, not the component instance. The language options' `onClick` handlers are all the same method, whereas having the binding inline like `onClick={this.selectLanguage.bind(this)}` would create a new function in memory for each option.

shaneriley:
@mitra42 not a problem. I've updated it to import the `languageConfig` and build it off of that.

packages/ia-components/sandbox/language-select/language-select.js
jbuckner: Oh wow that's interesting. Is this a common thing for React? It seems like you'd have to do that everywhere.

shaneriley: Yep. They have [a separate article in their docs](https://reactjs.org/docs/handling-events.html) addressing this. This isn't exclusive to React; it's been something to be aware of in most frameworks and libraries.

jbuckner: Ahhh. I don't remember having to do that with LitElement, but just found their docs.. it looks like Lit [takes care of that context](https://lit-element.polymer-project.org/guide/events#using-this-in-event-handlers) for you so the context is always the component itself.

jbuckner: I just saw in that article that you can do this and avoid the binding (it says experimental so it might not work):
```
class LoggingButton extends React.Component {
  // This syntax ensures `this` is bound within handleClick.
  // Warning: this is *experimental* syntax.
  handleClick = () => {
    console.log('this is:', this);
  }

  render() {
    return (
      <button onClick={this.handleClick}>
        Click me
      </button>
    );
  }
}
```

mitra42:
Thanks - I'll grab the code and try and integrate on the plane.

mitra42:
Thanks for this @shaneriley - I had to make some minor tweaks - just to do with the other CSS on the page, and commented on the code under the "changed files" section (for some reason they don't appear here). 

The only thing I've been unable to fix myself is the wrapping of the dropdown. If you look at https://www-dweb-mirror.dev.archive.org/details/local on Firefox it looks great, but on Chrome some of the lines in the drop-down wrap. (Flag on top, text underneath). 

Note - that URL is just my offline code running under Kubernetes to allow for testing.

shaneriley:
Adding `white-space: nowrap` to `.language_select li a` should fix it.

PR #383, opened 2020-02-11
Radio Transcript Searching

jbuckner:
**Description**

> Updates the radio player search to reconstruct the transcript on the client side. Previously we were getting a new transcript from the backend every time the user searched, but now we're building the transcript client-side.

**Technical**

> The search is somewhat complex because it has to rebuild the transcript with search results broken up by search results within the overall transcript. I tried to break up the search functionality as much as possible, add a bunch of tests and add a lot of comments to make it easier to digest.

**Testing**

> Run `yarn test` inside the `packages/radio-player` directory to run the tests. Also run `yarn start` and you can see the full demo app. Look at `search-handler.test.js` to see all of the test cases.

**Evidence**

> n/a


packages/radio-player/src/search-handler/search-index.ts
```diff
@@ -0,0 +1,139 @@
...
+   * This gets populated as part of the search index build. It merges all of the transcript
+   * entries together so we can search it as a single document instead of a bunch of
+   * individual entries. This allows searches to cross over transcript entries.
+   *
+   * It also indexes the entries so we can quickly look them up later.
+   *
+   * @private
+   * @memberof SearchIndex
+   */
+  private searchIndexCache: SearchIndexCache = new SearchIndexCache();
```
iisa: do we need to pass the transcript to it?

packages/radio-player/src/search-handler/search-index.ts
```diff
@@ -0,0 +1,139 @@
...
+   * @param {string} term
+   * @returns {Range[]}
+   * @memberof SearchIndex
+   */
+  getSearchRanges(term: string): Range[] {
+    const ranges: Range[] = [];
+    const termLowerCased: string = term.toLowerCase();
+
+    let index = -1;
+    /* eslint-disable-next-line no-cond-assign */
```
iisa: instead of disabling, can we just push a `new Range` into ranges?

packages/radio-player/src/search-handler/search-index.ts
```diff
@@ -0,0 +1,139 @@
...
+   *
+   * @private
+   * @memberof SearchHandler
+   */
+  private buildIndex(transcriptConfig: TranscriptConfig): void {
+    let startIndex = 0;
+    const transcriptEntryRanges: TranscriptEntryRange[] = [];
+    let mergedTranscript = '';
+
+    transcriptConfig.entries.forEach((entry: TranscriptEntryConfig) => {
```
iisa: 🍗

packages/radio-player/src/search-handler/search-handler.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   * Search the full transcript and split up by search results and non-results. For instance,
+   * if the full transcript is `foo bar baz boop bump snap pop` and you search for `bump`,
+   * you'll get an array of 3 results back:
+   * 1. `foo bar baz boop `
+   * 2. `bump` <-- the match
+   * 3. ` snap pop`
+   *
+   * This is helpful when rebuilding the transcript later to be able to identify search results.
```
iisa: cool

packages/radio-player/src/search-handler/search-handler.ts
```diff
@@ -0,0 +1,147 @@
...
+      const nonResultText = mergedTranscript.substring(startIndex, range.startIndex);
+      const resultText = mergedTranscript.substring(range.startIndex, nextStart);
+      const nonResultRange = new Range(startIndex, range.startIndex - 1);
+      const nonResultEntry = new SearchResult(nonResultRange, nonResultText, false);
+      const searchResultRange = new Range(range.startIndex, nextStart - 1);
+      const searchResultEntry = new SearchResult(searchResultRange, resultText, true);
+      transcriptEntries.push(nonResultEntry);
+      transcriptEntries.push(searchResultEntry);
+      startIndex = nextStart;
+    });
```
iisa: I might just need a walk through this part with you.

packages/radio-player/demo/radio-player-controller.ts
```diff
@@ -48,15 +40,54 @@ export default class RadioPlayerController extends LitElement {
...
+  async setup(): Promise<any> {
+    await this.loadItemMetadata();
+    await this.fetchTranscript();
+
+    const searchParams = new URLSearchParams(window.location.search);
+    const searchTerm = searchParams.get('q');
+    const startTime = searchParams.get('start');
+
+    if (searchTerm && this.radioPlayer) {
+      this.radioPlayer.searchTerm = searchTerm;
```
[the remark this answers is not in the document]
jbuckner: In practice, no, but this is TypeScript being strict about the possibility that it might not. `this.radioPlayer` is defined this way:
```
  private get radioPlayer(): RadioPlayer | null {
    return this.shadowRoot ? (this.shadowRoot.querySelector('radio-player') as RadioPlayer) : null;
  }
```
Since it's possible for there to not be a `shadowRoot`, as well as the possibility for that `shadowRoot.querySelector()` to return nothing, you have to assume that it's possible it won't exist.

In TypeScript 3.7, they added this super nice feature called Optional Chaining that would've let me do this:
```
private get radioPlayer(): RadioPlayer | null {
  return this.shadowRoot?.querySelector('radio-player') as RadioPlayer;
}
...
if (searchTerm) {
  this.radioPlayer?.searchTerm = searchTerm;
}
```

packages/radio-player/demo/radio-player-controller.ts
```diff
@@ -48,15 +40,54 @@ export default class RadioPlayerController extends LitElement {
     `;
   }
 
-  async loadItemMetadata() {
+  /* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
jbuckner: In general, I like this eslint rule because it enforces typing wherever possible. In this case, I'm dealing with the generic metadata JSON blob instead of a proper model so I have to use `any`.

This is a case where I'd like to flesh out our data models more. I'd like to come up with a full on data model for our metadata and pass that around instead of the JSON blob. This would let me get rid of this eslint disabling.

packages/radio-player/src/search-handler/search-index.ts
jbuckner: No, in this case, it just creates an empty cache. I did this so `searchIndexCache` wasn't an optional (and having to deal with optionals all over the place). The `searchIndexCache` gets recreated with a real cache as soon as the initialization finishes so this is just a placeholder until that happens.

packages/radio-player/demo/radio-player-controller.ts
```diff
@@ -48,15 +40,54 @@ export default class RadioPlayerController extends LitElement {
...
+          entry.id,
+          entry.start,
+          entry.end,
+          entry.text,
+          entry.is_music,
+          entry.search_match_index,
+        ),
+    );
+
+    this.transcriptConfig = new TranscriptConfig(transcriptEntries);
```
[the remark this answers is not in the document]
jbuckner: `transcriptEntries` is an array of each line of the transcript. Put together, they form the "transcript", which gets store in the `entries` property of the `TranscriptConfig` object. At it's most basic level, it just contains an array of entries, but it also contains some useful additions like extracting the search results from the transcript. ie:

```
const transcript = TranscriptConfig; // some TranscriptConfig object
const entries = transcript.entries // [TranscriptEntryConfig1, TranscriptEntryConfig2, TranscriptEntryConfig3 ...]
const searchResults = transcript.searchResults // [TranscriptEntryConfig2] (only the search results)
```

So here I'm creating a new `TranscriptConfig` object with the entries from `transcriptEntries`

packages/radio-player/src/search-handler/search-index.ts
jbuckner: Hmm maybe? I _think_ it was complaining about the `index = this.searchIndexCache.me...` in the `while` loop, not so much the `push(newRange)` part.

I'm not sure how to make this loop more readable, but definitely up for it.

packages/radio-player/src/search-handler/search-handler.ts
jbuckner: I cleaned up this part and added some documentation as to what's going on through it. Take another look and let me know if it's more clear now.

packages/radio-player/src/search-handler/search-index.ts
jbuckner: 😋

PR #385, opened 2020-02-12
385 mitra language seleect

mitra42:
Tweaks some of the CSS in @shaneriley 's language-select to make it work in dweb-mirror (its intended use case)


mitra42:
Note - it only changes 2 files, the other commits seem to be an artifact of how previous merges were rebased yesterday by @iisa  I'll look at getting them out.

packages/ia-components/sandbox/language-select/language-select.less
```diff
@@ -41,10 +41,15 @@
     &.visible {
       display: block;
     }
+    // It needs a more specific classname else the body.navia .navbar #nav-dweb ul { display:closed } beats the one below
```
shaneriley: Is the actual property value you're overriding `display: block`? `closed` isn't a property value.

mitra42: Sorry its display:inline that it overrides, I've pushed an update

shaneriley:
This MR makes changes to files that are no longer in master. I think we should create a new PR to add a 'visible'/'notvisible' class toggle in the component and have any CSS overrides for DWeb's implementation of the component exist outside of this base component's styles. Does that sound like it will work @mitra42 ?

mitra42:
Hmm - the actual file changes are just files in Master, @iisa and @bfalling and I had a chat about the problem of old changes getting carried over because of the squash merges, being done - with the conclusion that squash merges shouldnt be done on long-running branches.

mitra42:
Yes, new PR is fine by me ... do you want to do that because if I understand your last comment you want to do it slightly different from how I did. 
As to where the css changes should go, I'm not sure - there is a tension between having full encapsulation of local styles relating to the component, OR having long (full path) CSS that allows for it to be changed, so in this case if we want encapsulation I think we need the `!important` to make sure that the LI & UL aren't subject to the default designs of whatever is enclosing it.

I think we also need the deletion of the border, because as it is now it assumes the top level of the unexpanded drop-down is stand-alone, while in practice its always going to be inside something defined at the next level up - in this case the borders on the different nav buttons. 

Happy to discuss ....

PR #333, opened 2019-10-29
Creates ia-components release script

shaneriley:
**Description**

Adds version and postversion scripts to make tagging releases in GitHub easier.
Closes #332 . 

**Technical**

Version number should first be updated in package.json. Once updated, running `npm run version` will write a release message placeholder to CHANGELOG.md. Replace with your release message, save CHANGELOG.md, then press a key at the prompt. A browser window should open with the release info prefilled, the URL copied to the clipboard, and rendered to the terminal output.

**Testing**

Run `npm run version`

PR #392, opened 2020-02-19
Mobile topnav

shaneriley:
**Description**

Finishes mobile topnav components.

**Technical**

These styles are for mobile viewport widths only. Desktop viewport considerations will be added later, turning this component into an `<ia-nav>` component instead.

**Testing**

* `cd packages/mobile-top-nav`
* `yarn`
* `yarn start`

Some unit tests are also written.

* `yarn test`


packages/mobile-top-nav/src/css/media-menu.js
```diff
@@ -0,0 +1,69 @@
+import { css } from 'lit-element';
+
+export default () => {
```
jbuckner: How come you have all of these css blocks as separate files instead of part of their component?

packages/mobile-top-nav/src/data/menus.js
```diff
@@ -0,0 +1,337 @@
+const texts = {
+  heading: 'Books',
+  iconLinks: [{
+    title: 'Books to Borrow',
+    icon: 'https://archive.org/images/book-lend.png',
+    url: 'https://archive.org/details/inlibrary?sort=-publicdate',
```
jbuckner: Two thoughts:
1. I would package the icons in the repo so we're not relying on the production web server to serve up the icons.
2. I don't think we can use absolute URLs here, otherwise we can't use it in a review app.

packages/mobile-top-nav/README.md
```diff
@@ -22,65 +23,55 @@ export default IAMobileTopNav;
...
-    --trackBorder: 1px solid black;
-    --trackBorderRadius: 5px;
-    --trackHeight: 10px;
-    --thumbDiameter: 20px;
-    --scrubberBarHeight: 20px;
-    --thumbBorderRadius: 10px;
-    --webkitThumbTopMargin: -6px;
+  /* Defaults */
+  topnav-element {
+    --white: #fff;
```
jbuckner: Are your CSS variables named `white` and `grey20`? You want them to be more semantically meaningful like `titleColor` and `bodyColor` that can then be set to `#fff` or `#333`.

jbuckner:
Since this is such a big change to the base branch, I would just open this MR straight to `master` instead of `feature-mobile-topnav`.. it’s kind of hard to review this since the base branch wasn’t really ever done.

packages/mobile-top-nav/src/css/media-menu.js
shaneriley: I still like the idea of separating concerns. Going this route, if there are general styles that you want to include on multiple components, you could then

```
static get styles() {
  return [baseCss, cascadeCss];
}
```

packages/mobile-top-nav/src/data/menus.js
shaneriley: Yeah, there's a few spots like this. Ultimately, this menus file will be replaced with a fetch for the actual data to render the subnavs, so I figured in the short term this would be the least amount of cleanup.

packages/mobile-top-nav/README.md
shaneriley: I think Kanchan was modeling the variable names after the existing Less names in Petabox.

packages/mobile-top-nav/src/css/media-menu.js
jbuckner: Hrm I think this adds an unnecessary layer of indirection and is not really a separation of concerns. The WebComponent is a self-contained thing: markup, CSS, and logic. One of its huge benefits is to be able to work on it in isolation without jumping around through different files. If we split it up, we're now maintaining two parallel directories of CSS and Components and that is going to be a bear to manage.

I also don't think you should import external CSS's into your component. The style in your component should be completely self-contained. You quickly begin to run into the same problem with old school CSS where you become afraid to touch your shared stylesheet because you don't know what components you're going to affect. 

If you have customization points, they should be done with CSS Variables so the component is only exposing what it knows about.

packages/mobile-top-nav/src/assets/img/icon_web.js
```diff
@@ -0,0 +1,43 @@
+import { html } from 'lit-element';
+
+export default (fillColor) => (
```
jbuckner: Can you do these like you did with the Search and User icons below where you use the CSS class to change its styling?

packages/mobile-top-nav/src/assets/img/media-menu-images.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable */
 import { LitElement, html } from 'lit-element';
+import * as icons from './icons.js';
 
 class mediaMenuImage extends LitElement {
```
jbuckner: What is this being used for? I don't see it in the diff

packages/mobile-top-nav/src/assets/img/search.js
```diff
@@ -1,15 +1,35 @@
...
+  }
 
-class Search extends LitElement {
   static get properties() {
-    return {
-      colour: { type: String }
-    };
+    const superclassProps = super.properties;
+    return Object.assign({
+      width: { type: String },
```
jbuckner: Why is this a string?

packages/mobile-top-nav/src/assets/img/search.js
```diff
@@ -1,15 +1,35 @@
...
+    this.width = "27";
+    this.height = Math.ceil((+this.width) * 1.11);
+  }
 
-class Search extends LitElement {
   static get properties() {
-    return {
-      colour: { type: String }
-    };
+    const superclassProps = super.properties;
```
jbuckner: Are you doing anything with the `superclassProps`?

packages/mobile-top-nav/src/media-menu.js
```diff
@@ -129,6 +130,7 @@ class MediaMenu extends LitElement {
         <div class="menu-group">
           ${this.mediaMenuOptionsTemplate}
           <media-slider
+            selectedMenuOption=${this.selectedMenuOption}
```
jbuckner: I think you want to prefix this with a dot to bind it to the property, ie:
`.selectedMenuOption=${this.selectedMenuOption}`

https://lit-element.polymer-project.org/guide/templates#bind-properties-to-templated-elements

packages/mobile-top-nav/src/assets/img/search.js
shaneriley: I set it up as a string because the attribute value comes through as a string, not thinking about how LitElement will attempt to coerce it to the type specified.

packages/mobile-top-nav/src/assets/img/media-menu-images.js
shaneriley: It's being rendered in `<media-menu>`

packages/mobile-top-nav/src/assets/img/icon_web.js
shaneriley: Yeah, those can be converted to components instead. I was at a point where I was considering generating an SVG use map instead, but I don't think those are accessible outside of the topnav element's shadowRoot so it wouldn't be useful to do so.

packages/mobile-top-nav/src/assets/img/search.js
shaneriley: Nope. Didn't end up using them.

packages/mobile-top-nav/src/media-menu.js
shaneriley: That's probably something in need of updating across the components, since these aren't HTML attributes that need to be exposed.

packages/mobile-top-nav/src/media-slider.js
```diff
@@ -1,14 +1,73 @@
...
   }
 
+  shouldUpdate() {
+    if (menus[this.selectedMenuOption]) {
+      this.links = menus[this.selectedMenuOption];
+    }
+    return true;
+  }
+
+  renderIconLinks() {
```
jbuckner: For getters like this you can use this syntax:
```
get iconLinks() {
  return html`...`;
}
```
then in your `render()` method:
```
return html`
  ${this.iconLinks}
`;
```

packages/mobile-top-nav/src/media-slider.js
```diff
@@ -1,14 +1,73 @@
...
+  constructor() {
+    super();
+    const defaultLinks = { iconLinks: [], featuredLinks: [], links: [] };
+
+    this.mediaSliderOpen = false;
+    this.mediaSliderAnimate = false;
+    this.selectedMenuOption = 'texts';
+
+    // Begin properties not monitored by LitElement
+    this.links = menus[this.selectedMenuOption] || defaultLinks;
```
jbuckner: I would make `links` a reactive property since there are several dependencies down the line like in `renderLinks` and `renderIconLinks`
```
get links() {
  return menus[this.selectedMenuOption] || this.defaultLinks;
}
```
I think this would let you get rid of that `shouldUpdate` lifecycle event below.

packages/mobile-top-nav/src/media-slider.js
```diff
@@ -1,14 +1,73 @@
...
+      <div class="icon-links">
+        ${this.renderIconLinks()}
+      </div>
+      <ul>
+        ${this.renderLinks('featuredLinks')}
+      </ul>
+      <h4>Top</h4>
+      <ul>
+        ${this.renderLinks('links')}
+      </ul>
```
jbuckner: I would structure your `renderSubnav` a bit differently and may even consider pulling the subnav out to be its own component. It looks like there's a decent amount of logic going on in it.
```
get subnav() {
  const template = this.templates[this.selectedMenuOption];
  return template ? template() : this.defaultSubnav;
}

get defaultSubnav() {
    return html`
       <h3>${this.links.heading}</h3>
       <div class="icon-links">
       ...
       ${this.featuredLinks} <--- so it doesn't get executed immediately, keeps it reactive
       ${this.nonFeaturedLinks}
    `;
}

get featuredLinks() {
  return this.renderLinks('featuredLinks');
}

get nonFeaturedLinks() {
  return this.renderLinks('links');
}

renderLinks(category) {
   return this.links[category].map((link) => html`<li><a href="${link.url}">${link.title}</a></li>`);
}
```

packages/mobile-top-nav/src/media-slider.js
shaneriley: The `shouldUpdate` behaves slightly different than an either/or. When the subnav is closed, if we use a getter like above, the menu is replaced with an empty one and it's visible as it animates out. Instead, I'm leaving the current menu in place when `selectedMenuOption` is unset.

jbuckner: Oh I see.. hmm since that has dependent logic on the changing of `selectedMenuOption`, it might be a good candidate for the `updated(changedProperties)` callback instead of `shouldUpdate`. `updated()` is the more common place to handle property-dependent changes. I would probably still make `links` a property, though so any listeners can update when it changes. 

ie:

```
updated(changedProperties) {
  if (changedProperties.has('selectedMenuOption')) {
    this.handleSelectedMenuOptionChange();
  }
}

handleSelectedMenuOptionChange() {
  if (menus[this.selectedMenuOption]) {
    this.links = menus[this.selectedMenuOption];
  }
}
```

shaneriley:
Going to merge this onto the original feature branch and open a new PR

PR #396, opened 2020-02-25
Radio Player: Don't Display the Waveform if one is not Available

jbuckner:
**Description**

> If the radio program does not have a waveform image, don't display the waveform component.

**Technical**

> n/a

**Testing**

> Go to https://58-review-radio-play-lpp4ra.archive.org/details/BBC_Radio_5_Live_20191029_000000 and you shouldn't see a broken waveform

**Evidence**

> 
Screen Shot 2020-02-25 at 10 22 01 AM

Screen Shot 2020-02-25 at 10 22 14 AM

### DIFF
--- a/packages/radio-player/src/radio-player.ts
+++ b/packages/radio-player/src/radio-player.ts
@@ -274,7 +274,7 @@ export default class RadioPlayer extends LitElement {
    * @memberof RadioPlayer
    */
   private get waveFormProgressTemplate(): TemplateResult {
-    return html`
+    return this.waveformUrl ? html`
       <waveform-progress
         interactive="true"
         .waveformUrl=${this.waveformUrl}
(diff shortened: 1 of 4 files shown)


packages/radio-player/src/radio-player.ts
```diff
@@ -274,15 +274,15 @@ export default class RadioPlayer extends LitElement {
...
+    return this.waveformUrl ? html`
       <waveform-progress
         interactive="true"
         .waveformUrl=${this.waveformUrl}
         .percentComplete=${this.percentComplete}
         @valuechange=${this.valueChangedFromScrub}
       >
       </waveform-progress>
-    `;
+    ` : html``;
```
iisa: is this the way we should be returning "nothing"? In React, we return a `null` and it doesn't add excess markup to actual DOM.

jbuckner: I changed it to return `undefined` if there's no template. That does the trick! Good call.

PR #397, opened 2020-02-25
Playback Controls Mute Fix

jbuckner:
**Description**

> Fixes an issue where the player doesn't mute.

**Technical**

> If you try to set the volume to 0, some responders don't respond since 0 is falsey in javascript. Instead, just turn the volume to a very low non-0 value.

**Testing**

> Checkout this branch, `cd packages/radio-player`, `yarn install`, `yarn start`.

**Evidence**

> n/a

### DIFF
--- a/packages/playback-controls/src/playback-controls.ts
+++ b/packages/playback-controls/src/playback-controls.ts
@@ -16,11 +16,13 @@ import volumeMuteImage from './assets/img/volume/volume-mute';
 
 @customElement('playback-controls')
 export default class PlaybackControls extends LitElement {
-  @property({ type: PlaybackMode }) playbackMode = PlaybackMode.paused;
+  @property({ type: Number }) playbackMode = PlaybackMode.paused;
 
   @property({ type: Number }) playbackRate = 1;
 
   @property({ type: Number }) volume = 1;
+
+  private readonly minimumVolume = 0.00001;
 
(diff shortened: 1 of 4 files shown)


packages/playback-controls/src/playback-controls.ts
```diff
@@ -102,12 +108,21 @@ export default class PlaybackControls extends LitElement {
   }
 
   handleVolumeChange() {
-    if (this.volume === 1) {
-      this.volume = 0;
+    let newVolume = this.volume;
+
+    if (newVolume === 1) {
+      // we cannot use a volume of 0 since it becomes "falsey" and causes it to not be set
```
iisa: So, you are starting at 1, what is max? 10?  What is mute, 1?

iisa: what is the "thing" that reads it as falsey?

jbuckner: The range is 0 to 1, but we're sort of shortcutting the `0` to `0.00001` so it doesn't become falsey.

jbuckner: 🤦‍♂ I just found the thing intercepting the falsey value and stopping. Going to update this MR.

jbuckner:
There's a much simpler fix for this 😅

PR #399, opened 2020-02-26
Radio Player Mute Fix

jbuckner:
**Description**

> Fix the mute setting in the volume control.

**Technical**

> Since the volume can be `0`, we have to specifically check for `undefined` instead of just falsey.

**Testing**

1. With an admin archive.org email user
2. Visit https://58-review-radio-play-lpp4ra.archive.org/details/VOA_Global_English_20170803_190000?start=0
3. Play and make sure you can mute the volume.

**Evidence**

> n/a

### DIFF
--- a/packages/radio-player/src/radio-player.ts
+++ b/packages/radio-player/src/radio-player.ts
@@ -716,7 +716,7 @@ export default class RadioPlayer extends LitElement {
     if (detail.volume === undefined) {
       return;
     }
-    this.volume = e.detail.volume;
+    this.volume = e.detail.volume
   }
 
   /**
--- a/packages/radio-player/src/radio-player.ts
+++ b/packages/radio-player/src/radio-player.ts
@@ -716,7 +716,7 @@ export default class RadioPlayer extends LitElement {
     if (detail.volume === undefined) {
       return;
     }
-    this.volume = e.detail.volume
+    this.volume = e.detail.volume;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/radio-player/src/radio-player.ts
```diff
@@ -713,7 +713,7 @@ export default class RadioPlayer extends LitElement {
    */
   private volumeChanged(e: CustomEvent): void {
     const detail = e.detail || {};
-    if (!detail.volume) {
+    if (detail.volume === undefined) {
```
iisa: 👏

packages/radio-player/src/radio-player.ts
```diff
@@ -716,7 +716,7 @@ export default class RadioPlayer extends LitElement {
     if (detail.volume === undefined) {
       return;
     }
-    this.volume = e.detail.volume;
```
shaneriley: You can `git commit --amend` changes to a commit that hasn't been pushed up to origin yet without a force push. Then you'll be able to eliminate the 'bump' commits.

jbuckner: Do those commits count as individual commits? I did that because `lerna` seems to count the number of commits to generate the canary version and if you try to re-publish the same version, it fails.

packages/radio-player/src/radio-player.ts
jbuckner: 🤦‍♂ 😅

PR #402, opened 2020-03-08
aligned input area to button

anishagg17:
**Description**

aligned input area to button

**Technical**

added a new class for styling 

**Evidence**

before:

Screenshot 2020-03-08 at 3 01 59 PM

after:

Screenshot 2020-03-08 at 3 01 36 PM


shaneriley:
There is a much simpler fix to this alignment by changing the `vertical-align` property on the `.btn` class definition in https://github.com/internetarchive/iaux/blob/master/packages/ia-components/.storybook/archive-cf7f2eh.less#L1786 to 'baseline'. This way, we don't have CSS in our component that is purely used in the presentation of Storybook controls.

Issue #222, opened 2019-07-01
Stop autoplaying the youtube player after changing channels


koderjoker:
Task Details
Currently, our youtube player autoplays the selected track after switching channels. Our aim is to make sure the selected track doesn't autoplay after doing so.

Batey96:
Hey Koder,is it similar to my PR #128?

koderjoker:
Nope, since this one deals with the youtube component (youtube-player.jsx). 

The music player has seen many changes since then.

iisa:
done

PR #403, opened 2020-03-08
fixed contrast ratio, accessibility issues

anishagg17:
**Description**

fixed contrast ratio, accessibility issues

**Technical**

fixes contrast ratio in language-select by using a more brighter colour
fixes accessibility issue in iajs component by adding aria-lable

**Evidence**

Screenshot 2020-03-08 at 3 30 23 PM

Screenshot 2020-03-08 at 3 29 50 PM

### DIFF
--- a/packages/ia-components/sandbox/iajs-test-component/index.jsx
+++ b/packages/ia-components/sandbox/iajs-test-component/index.jsx
@@ -32,17 +32,18 @@ export default class extends Component {
   fetchMetadata(identifier) {
     let metadataService = new MetadataService()
     // TODO-ERRORS Note this has an unhandled error if MetadataService.get fails
-    metadataService.get({identifier: identifier}).then(metadata => {
-      this.setState({metadata: metadata.data})
+    metadataService.get({ identifier: identifier }).then(metadata => {
+      this.setState({ metadata: metadata.data })
     })
   }
 
-  render () {
+  render() {
     return (
       <div className="iajs-test-component">
         <h5>My identifier is {this.props.identifier}</h5>
         <img
-          style={{maxWidth: '300px', height: 'auto'}}
+          aria-label="iajs-test-component"
+          style={{ maxWidth: '300px', height: 'auto' }}
           src={`https://archive.org/services/img/${this.props.identifier}`}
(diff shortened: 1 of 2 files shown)


packages/ia-components/sandbox/iajs-test-component/index.jsx
```diff
@@ -32,17 +32,18 @@ export default class extends Component {
...
   }
 
-  render () {
+  render() {
     return (
       <div className="iajs-test-component">
         <h5>My identifier is {this.props.identifier}</h5>
         <img
-          style={{maxWidth: '300px', height: 'auto'}}
+          aria-label="iajs-test-component"
```
shaneriley: This image is entirely decorative, and as such it makes more sense to simply add an empty `alt` attribute to conform to a11y standards as described in https://www.w3.org/WAI/tutorials/images/decorative/

packages/ia-components/sandbox/language-select/language-select.less
```diff
@@ -1,5 +1,5 @@
 @dropdown_background_color: #222222;
-@text_color: #848484;
+@text_color: #b9b9b9;
```
shaneriley: I've put in a request to have this color change approved by our design team before approving. Will let you know with a follow-up comment if this change is acceptable.

packages/ia-components/sandbox/iajs-test-component/index.jsx
shaneriley: The alt attribute should be empty, not populated with the parent div's class value.

packages/ia-components/sandbox/language-select/language-select.less
shaneriley: Ok, from the design team: "We've been moving to #999999 for these. It's AA compliant for small text against the #222 background, and it's easier to differentiate from the #fff active text color." Updating this to #999999 and setting the image alt attribute to "" will make this ready to merge.

anishagg17:
@shaneriley I have made the requested changes

anishagg17:
@shaneriley I have made the requested changes

PR #404, opened 2020-03-09
Adds TrackedElement LitElement decorator and example usage

shaneriley:
**Description**

This creates a class decorator to make it easier to emit click and submit tracking events. In Petabox, these will then be passed on to analytics.js

**Technical**

See README for details on implementation

**Testing**

`yarn start` will open an interactive demo. `yarn test` will run unit tests.


packages/analytics-element/src/tracked-element.js
```diff
@@ -0,0 +1,27 @@
+import { LitElement } from 'lit-element';
+
+class TrackedElement extends LitElement {
```
jbuckner: This class strikes me more as a [Mixin](https://javascript.info/mixins) than a LitElement. By itself, it doesn't really do anything except add functionality to another class.

In general, I'm averse to making "base classes". Over time, base classes tend to envelope a lot more functionality than intended because they become a convenient place to put functionality.

----

I don't think it's a good idea to let the events bubble up through the elements. I think they should only go up one level, otherwise you end up with these scattershot events that anything will receive, even unintentionally, and create the potential for conflicting events.

For tracking through petabox, I would probably have a hierarchy like this:

- `petabox`:
  - `<my-custom-element>`
    - `<child-element>`

`child-element` dispatches events to `my-custom-element` without bubbling
`my-custom-element` re-dispatches events (if needed) to `petabox`
`petabox` has listeners for all of the events it is interested in and tracks them

shaneriley: I'm not understanding what converting this to a mixin will change. A mixin allows you to take multiple mixins and `Object.assign` them together onto an object prototype, but so does subclassing. As far as throwing random functionality in this class, that's more of a matter of discipline than approach.

Regarding event bubbling, when I tested setting `bubbles` to false, it still bubbled out to the outermost custom element, so I don't see how changing that property will change whether or not the event bubbles up to parent elements. Setting composed to false will restrict the event to the current shadowRoot, but that's not what we want in this situation.

Even if the `bubbles: false` value on the custom event prevented event bubbling, that means that every time we have nested children that we want to track events for, we need to create a means of re-dispatching the event with every element beyond the first, when what we really need is a way for any component, no matter how nested, to be able to broadcast out to the parent document that a trackable event has occurred. You can see this in action at https://git.archive.org/ia/petabox/commit/801435250481bb8d96270f242709f4cf43161449 If there's an implementation that is as easy as this without event bubbling, I'll update it, but I didn't see a way of doing so without manually chaining together method calls and event listeners.

jbuckner: Re: Mixin: Subclassing brings a lot of other baggage with it, disciplined or not. In this case, by subclassing this, you are also required to have LitElement. While the current use case is Lit, the functionality it provides is not really Lit-specific.

---

Re: Bubbling/composed: My general opinion of events is that they should only go one level up and be semantically meaningful for their context. This does add a requirement of "re-dispatching" but that should be up to the intermediary component to decide whether the call gets propogated.

---

Events should also be more specific than `trackClick`. Click tracking is an action that the consumer should decide. The events should be more "what" is actually happening... ie `media-button-clicked`, `hamburger-menu-clicked`. The consumer can then consume those events and decide what to do with them. 

---

The other thing I noticed in the analytics example you posted is you're relying on the sender to determine what the analytics message is. ie whatever event gets emitted, gets tracked verbatim. If this component gets used in a different context than petabox, we may not want to use the same tracking information.

I would instead listen for specific events and then tailor the analytic message for the event. ie in petabox:
```
html`
  <topnav-element 
    @media-button-clicked=${this.trackMediaButtonClicked}
    @hamburger-clicked=${this.hamburgerClicked}>`;

trackMediaButtonClicked(e) {
  // track it
}

hamburgerClicked(e) {
  // track it
}
```

shaneriley: Oh, I see what you mean now when it comes to subclassing. I'll switch it out with a mixin pattern.

Regarding event emission, what I was trying to do was allow for the same mechanism we already use in Petabox to emit events. We use these two attribute names to then bind either a click or submit event. Those handlers read their respective attribute values, split them at the pipe, and append them as query params to the URL for 0.gif. If we go with a different approach now, we're going to have two different sets of analytics code.

jbuckner: This is a different world than petabox. In petabox, the tracking is tightly bound to the environment because there was a pattern set that when you add the `data-event-*` attribute, your link magically gets tracked. Here we are developing standalone components that should be able to be used in any consumer. 

There is nothing inherent in the component itself that should be concerned with any tracking. The components should only be responsible for their own functionality and emitting events for the things that other consumers may be interested in.

If the consumer is interested in tracking some event that a component has emitted, that's fine, but the component shouldn't be dictating that.

I don't think this is setting a good pattern for building standalone components.

packages/analytics-element/src/tracked-element.js
```diff
@@ -0,0 +1,23 @@
...
+    const event = e.currentTarget.dataset.eventSubmitTracking;
+    this.dispatchEvent(new CustomEvent('trackSubmit', {
+      bubbles: true,
+      composed: true,
+      detail: {
+        event
+      },
+    }));
+  }
+};
```
iisa: a few questions, comments come to mind with this file, they may compound on each other.  Let me list as I think, perhaps we can solidify some of my thoughts together:
- seems like each method is the same inside..
- can we whitelist these events, do we even need a white list?
- I can see the value in an explicit list, that way we can catch all the various events - what will this list look like?
- for "generic" purposes, should we extract these event names as constants that are defaults but can be overridden?


archive.org events look like this: `Foo|Bar` - I don't think the `example-component` should know this syntax - we should abstract this up one level into a helper function of some kind inside the tracking functionality.  I think this will help generalize the behavior on the consumption level.

ex.
`this.trackEvent({ event: 'Submit', element: 'ExampleTracker', details: 'foo' })`
where track event uses archive.org specific transformer (as default or can take another transformer) to make string and then fire.

packages/analytics-element/src/example-component.js
```diff
@@ -0,0 +1,47 @@
...
+      <form action="" method="GET" @submit=${this.submit} data-event-submit-tracking="ExampleTracker|Submit">
+        <fieldset>
+          <button @click=${this.click} data-event-click-tracking="ExampleTracker|Click">Track click</button
+        </fieldset>
+      </form>
+    `;
+  }
+}
+
+Object.assign(ExampleComponent.prototype, TrackedElement);
```
iisa: I don't think we should bind trackers to the object.  They are pure functions, and don't need/modify state, can they just be used ad-hoc in the code?
Perhaps we need to set a flag instead to show in constructor that we are tracking these links? `this._sendsEvents` or something..  I really like immediately knowing if this component has a special side effect like this, (taking a page from @cdrini and his preference of immediate declarations).

PR #407, opened 2020-03-14
Upgrade Typescript Components to TS 3.8 #390

### LINKED ISSUE #390: Upgrade Typescript Components to TS 3.7
**Task Details**

> We are currently on Typescript 3.6 and with Typescript 3.7, we get optional chaining and null coalescing operators. The optional chaining in particular will allow us to reduce a lot of code from 
```
if (foo) {
  foo()
}
```
to
```
foo?()
```

**Verification**

> TypeScript is updated in package.json. Spots that can be optionally chained should be switched to use optional chaining. 

**Stakeholders**

> @jbuckner

lelouchB:
**Description**

Closes #390
 
**Technical**

In this PR, typescript version is being updated to 3.8,3, Optional Chaining will be implemented in the next PR @jbuckner 

**Testing**

**Evidence**

### DIFF
--- a/packages/expandable-search-bar/package.json
+++ b/packages/expandable-search-bar/package.json
@@ -61,7 +61,7 @@
     "karma-coverage": "^2.0.1",
     "lint-staged": "^8.0.0",
     "polymer-cli": "^1.9.11",
-    "typescript": "^3.6.2",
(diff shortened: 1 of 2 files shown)


lelouchB:
The error is because of travisci config file for which I have send #408

PR #422, opened 2020-03-25
Positions timestamp from right side of time column

shaneriley:
When timestamp exceeds 1 hour (01:00:00), in the demo and on archive.org the timestamp exceeds the width of the time column and interferes with the transcript column. The timestamp is now positioned from the right edge of the column, allowing it to grow from right to left, and the demo has been updated to provide an example radio item that exceeds an hour as well as an increase in the time column width to accommodate times that exceed 1 hour.

Note that once this is published, the width of the time column on archive.org's implementation will need to be increased in kind for this fix to fully take effect.

**Description**

Fixes [WEBDEV-3189](https://webarchive.jira.com/browse/WEBDEV-3189)

**Testing**

Unable to fully test until the transcript-view component is published and radio-player is updated to use the newer version.

### DIFF
--- a/packages/radio-player/demo/index.html
+++ b/packages/radio-player/demo/index.html
@@ -17,7 +17,7 @@
       color: white;
 
       --timeColor: white;
-      --timeColumnWidth: 3rem;
+      --timeColumnWidth: 4rem;
       --transcriptHeight: 400px;
 
       --autoScrollButtonFontColor: black;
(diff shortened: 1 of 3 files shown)


jbuckner:
This runs into a similar issue if the time it up near the top of the scroll section (it gets cut off on the left):

Screen Shot 2020-03-25 at 8 34 23 AM

In the original specs (see the [screenshot](https://webarchive.jira.com/browse/WEBDEV-2371)), the time display was a smaller font size so Jim and I were discussing just fixing this by adjusting the column width and setting a `--timeFontSizeCss` CSS variable to make the font smaller. Both can be done in petabox.

PR #423, opened 2020-04-01
Dispatch Events from the Playback Controls to the Radio Player Consumer

jbuckner:
**Description**

> We were dispatching events from several of the components like scrubber bar scrubs and transcript entry selections, but not the playback controls.

This adds some event dispatches and tests for them.

**Technical**

> It is mainly re-dispatching events for the consumer to handle or track.

**Testing**

> Nothing to test. There are unit tests verifying that the events are being emitted.

### DIFF
--- a/packages/radio-player/test/radio-player.test.js
+++ b/packages/radio-player/test/radio-player.test.js
@@ -1044,11 +1044,11 @@ describe('Radio Player', () => {
       `);
 
       const event = new CustomEvent('foo', {
-        detail: { searchResultIndex: 2 },
+        detail: { searchResultIndex: 3 },
       });
       setTimeout(() => { el.searchResultIndexChanged(event); });
       const response = await oneEvent(el, 'highlightedSearchResultChanged');
-      expect(response.detail.searchResultIndex).to.equal(2);
+      expect(response.detail.searchResultIndex).to.equal(3);
     });
   });
 });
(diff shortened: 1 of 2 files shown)


packages/radio-player/test/radio-player.test.js
```diff
@@ -866,4 +866,207 @@ describe('Radio Player', () => {
...
+
+      const event = new CustomEvent('foo', {
+        detail: { volume: 0.5 },
+      });
+      setTimeout(() => { el.volumeChanged(event); });
+      const response = await oneEvent(el, 'volumeChanged');
+      expect(response.detail.volume).to.equal(0.5);
+    });
+
+    it('emits jumpBackButtonPressed event', async () => {
```
shaneriley, line 897: **Suggestion**: For some of these, you can abstract the code to a generic function and just pass in the event name in test. So with something like:

```javascript
const testEventEmitter = (eventName) => {
  it(`emits ${eventName} event`, async () => {
    const el = await fixture(html`
      <radio-player></radio-player>
    `);

    setTimeout(() => { el.canplay(); });
    const response = await oneEvent(el, eventName);
    expect(response).to.exist;
  });
};
```

you could then write `['jumpBackButtonPressed', 'jumpForwardButtonPressed', ...].forEach(testEventEmitter);`

jbuckner: Inside the `setTimeout`, I'm doing different things for each event, ie `setTimeout(() => { el.canplay(); });` produces a `canplay` event. 

How would you do that with this approach?

shaneriley: Yeah, sorry. The actual code should call the method like `setTimeout(() => { el[eventName]() })`. Then you can convert a number of these to use this same pattern.

shaneriley: If the method names differ, you could also use an array of objects instead, so `{ name: 'jumpFOrwardButtonPressed', method: 'forwardButtonHandler' }`

PR #426, opened 2020-04-10
<ia-topnav>

shaneriley:
**Description**

Replaces the mobile-top-nav with a unified nav component.

**Technical**

No functionality changes. All changes are made in service of using media query breakpoints to change style of components to suit the viewport size.

**Testing**

Run the app locally with `yarn start` and view in various browsers and viewport widths.


packages/ia-topnav/src/ia-topnav.js
```diff
@@ -0,0 +1,237 @@
...
+    const closeLayerVisible = this.isMenuOpen;
+
+    return html`
+      <div class='topnav'>
+        <primary-nav
+          .config=${this.config}
+          .selectedMenuOption=${this.selectedMenuOption}
+          .mediaMenuOpen="${this.mediaMenuOpen}"
+          .searchMenuFade="${this.searchMenuFade}"
+          .searchMenuOpen="${this.searchMenuOpen}"
```
jbuckner, line 195: What do these `*Open` properties do?

packages/ia-topnav/src/ia-topnav.js
```diff
@@ -0,0 +1,237 @@
...
+  navSearch(e) {
+    this.searchSubmitted = true;
+    const { originalEvent, formEl } = e.detail;
+    const searchIn = document
+      .querySelector('ia-topnav')
+      .shadowRoot
+      .querySelector('search-menu')
+      .shadowRoot
+      .querySelector('[name=sin]:checked')
+      .value;
```
jbuckner, line 139: This is tunneling awfully deep into other components to get a value. Can you store the checked value as a property or emit an event from the child components when that thing changes so consumers like this component can be aware of these changes?

packages/ia-topnav/src/nav-search.js
```diff
@@ -0,0 +1,56 @@
...
+    this.config = {};
+    this.open = false;
+  }
+
+  search(e) {
+    this.trackSubmit(e);
+    this.dispatchEvent(new CustomEvent('navSearch', {
+      detail: {
+        originalEvent: e,
+        formEl: this.shadowRoot.querySelector('form'),
```
jbuckner, line 29: How come you're passing the whole form element up?

packages/ia-topnav/src/styles/dropdown-menu.js
```diff
@@ -0,0 +1,105 @@
+import { css } from 'lit-element';
+
+export default css`
+  :host {
+    --topOffset: -1000px;
```
jbuckner, line 5: What is this doing?

packages/ia-topnav/src/dropdown-menu.js
```diff
@@ -0,0 +1,68 @@
...
+    ));
+  }
+
+  get menuClass() {
+    if (this.open) {
+      return 'open';
+    }
+    if (this.animate) {
+      return 'closed';
+    }
```
jbuckner, line 41: Is `this.open` and `this.animate` being used to change the state of the menu like open or closed?

If so, it seems like you could get into a case here where `open` is `true` and `animate` is `true` and you might expecting it to be closed, but it wouldn't.

An alternative is to create methods to control that state. This would let you guarantee internally that you can't get into those states. eg:
```
// dropdown-menu.js
openMenu() {
  this.menuClass = 'open'
}

closeMenu() {
  this.menuClass = 'closed'
}
...

// consumer-component.js
const menu = this.shadowDom.querySelector('dropdown-menu');
menu.openMenu();
```

packages/ia-topnav/src/ia-topnav.js
```diff
@@ -0,0 +1,237 @@
...
+    this.searchMenuOpen = !this.searchMenuOpen;
+    if (this.searchMenuOpen) {
+      window.setTimeout(() => {
+        this
+          .shadowRoot
+          .querySelector('primary-nav')
+          .shadowRoot
+          .querySelector('nav-search')
+          .shadowRoot
+          .querySelector('.search-field').focus();
```
jbuckner, line 108: Similar comment about tunneling into components. If any of the child components change their component structure and change these, this will break.

As an alternative, you could call methods on child components to tell their children to give the search field focus. This allows components to only be responsible for their immediate children.
ie:
```
// ia-topnav.js
this.shadowRoot.querySelector('primary-nav').focusSearch();

// primary-nav.js
focusSearch() {
  this.shadowRoot.querySelector('nav-search').focusSearch();
}

// nav-search.js
focusSearch() {
  this.shadowRoot.querySelector('.search-field').focusSearch();
}
```

packages/ia-topnav/src/media-menu.js
```diff
@@ -0,0 +1,127 @@
...
+    }) => {
+      const selected = this.selectedMenuOption === menu;
+      return html`
+        <media-button
+          .config=${this.config}
+          .icon=${icon}
+          .href=${href}
+          .followable=${followable}
+          .label=${label}
+          mediatype=${menu}
```
jbuckner, line 96: Dot needed here?

jbuckner:
_Suggestion_ All of the components are at the top-level so looking in the directory, it's hard to tell what the hierarchy is. I would combine some of the related components into subdirectories to make it more obvious how things are related.

jbuckner:
Reviewing components with separate CSS makes it much more difficult to get a picture of what's going on.

packages/ia-topnav/src/styles/desktop-subnav.js
```diff
@@ -0,0 +1,43 @@
...
+    height: 16px;
+    vertical-align: -4px;
+    fill: #f00;
+  }
+
+  @media (max-width: 889px) {
+    ul {
+      display: none;
+    }
+  }
```
jbuckner, line 42: Is this causing the desktop subnav to hide under 890px? If so, I would place this responsibility on its parent component. This change can affect layout in the parent so it should be aware that this is going to happen.

packages/ia-topnav/src/styles/nav-search.js
```diff
@@ -0,0 +1,134 @@
...
+      opacity: 1;
+    }
+  }
+  .fade-in {
+    animation: fade-in .2s forwards;
+  }
+
+  @media (min-width: 890px) {
+    :host {
+      float: right;
```
jbuckner, line 91: Does this cause layout changes in the parent? If so, I would move the responsibility up a level.

packages/ia-topnav/src/dropdown-menu.js
bfalling: This maybe be a totally different case, but I'm also thinking of lots of places in the merged mobile nav code where the second conditional was: `if (!this.open && this.animate) {`.

packages/ia-topnav/src/ia-topnav.js
bfalling: Law of Demeter: "Only talk to your friends." (not to friends of friends)

packages/ia-topnav/src/user-menu.js
```diff
@@ -0,0 +1,39 @@
...
+    const props = {
+      ...DropdownMenu.properties,
+      username: { type: String },
+    };
+    return props;
+  }
+
+  constructor() {
+    super();
+    this.username = 'USERNAME';
```
bfalling, line 20: QUESTION: Will this value ever get used anywhere? (Is there a need to initialize to this value?)

packages/ia-topnav/src/styles/ia-topnav.js
```diff
@@ -0,0 +1,42 @@
...
+    --grey13: #222;
+    --grey20: #333;
+    --grey28: #474747;
+    --grey999: #999;
+    --grey80: #ccc;
+    --black: #000;
+    --link-color: #428bca;
+    --subnavLinkColor: #aaa;
+    --primary-text-color: var(--white);
+    --activeColor: var(--white);
```
bfalling, line 15: I know the use of CSS variables for colors is a hot topic. It was for my past teams as well until we realized that there are really *two* needs for variables:

1. We never want raw hex numbers for colors, cuz they're meaningless and error prone, so we need good descriptive names, e.g. "scarlet" and "grey999". These variable values should never change.
2. We want variables to assign colors to meaningful contexts, e.g. "defaultLinkColor" or "topNavBackground". These values change as design demands change.

The key decoupling was that variables of type 2 should be set to variables of type 1. So setting `--white: #fff` and then `--activeColor: var(--white)` is a perfect example of this.

I've found it helpful to group the two kinds of variables together, so it's clear which ones are "color constants" and which are "design assignments".

bfalling:
QUESTION: In the original merge request, `toSnakeCase.js` seems like it's 1) converting to camel case, not snake case, and 2) able to get into infinite recursion. I might be wrong, but please do check it out and fix or remove if not used.

packages/ia-topnav/src/more-slider.js
```diff
@@ -1,7 +1,8 @@
-import { html, css } from 'lit-element';
+import { html } from 'lit-element';
 import TrackedElement from './tracked-element';
 import { more as moreMenu } from './data/menus';
 import toSnakeCase from './lib/toSnakeCase';
```
bfalling, line 4: QUESTION: In the original merge request, `toSnakeCase.js` seemed like it's 1) converting to camel case, not snake case, and 2) able to get into infinite recursion. I might be wrong, but please do check it out and fix or remove if not used.

packages/ia-topnav/src/ia-topnav.js
```diff
@@ -145,55 +149,69 @@ export default class TopnavElement extends LitElement {
...
           .config=${this.config}
-          ?mediaMenuOpen="${this.mediaMenuOpen}"
-          ?searchMenuFade="${this.searchMenuFade}"
-          ?searchMenuOpen="${this.searchMenuOpen}"
-          ?userMenuOpen="${this.userMenuOpen}"
+          .selectedMenuOption=${this.selectedMenuOption}
+          .mediaMenuOpen="${this.mediaMenuOpen}"
+          .searchMenuFade="${this.searchMenuFade}"
+          .searchMenuOpen="${this.searchMenuOpen}"
+          .userMenuOpen="${this.userMenuOpen}"
```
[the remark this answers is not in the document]
shaneriley: Looks like nothing since the changes.

packages/ia-topnav/src/ia-topnav.js
shaneriley: Determine opened state of various menus.

packages/ia-topnav/src/ia-topnav.js
shaneriley: Yep. One of those bits of code that I put in initially while I rearranged elements and meant to go back to.

packages/ia-topnav/src/nav-search.js
shaneriley: Whole thing's been refactored now.

packages/ia-topnav/src/styles/dropdown-menu.js
shaneriley: Allowing for the quick change of the starting top offset. Since it's used in two spots, this gets rid of what Brenton typically points out to be a "magic constant".

packages/ia-topnav/src/ia-topnav.js
shaneriley: Your approach also breaks if the child components change. If, for example, the `<nav-search>` element is not rendered within `<primary-nav>`, a TypeError will be thrown.

I'll move it to a prop and pass it down through the components. This pattern is reminding me why I loved Redux so much.

jbuckner: But if you remove `<nav-search>` from `<primary-nav>`, `<primary-nav>` is aware of that change since it's one of its own children and can be updated accordingly. It may change its behavior when `focusSearch()` is called, but `ia-topnav` doesn't need to be concerned about that.

It's about defining the API of each component and limiting your interaction to that.

shaneriley: Switched to an `updated` method and a check for whether the component's `open`.

packages/ia-topnav/src/media-menu.js
shaneriley: Yep. Not an HTML attribute.

packages/ia-topnav/src/dropdown-menu.js
shaneriley: The `animate` properties are used in conjunction with the `open` properties to toggle animation. When the element is first rendered, there's an initial class added to the element that prevents the CSS properties from transitioning into initial position. Once open state is changed, the animate property is set to true and remains true, since from that point forward the animation should move forward and in reverse with each toggle. This is a pattern that is carried over from the initial implementation of the mobile nav.

I'll look into refactoring the behavior project-wide, but it's going to be lower on the priorities. There's an outstanding issue of the dropdowns animating out of the way when you dynamically change the viewport width between breakpoints that will touch these properties and class changes.

packages/ia-topnav/src/more-slider.js
shaneriley: Yep. Should be `toSentenceCase`. I don't see how there'd be infinite recursion, since it creates an array at space characters and reduces the number of space characters by 1 with each new recursive call.

packages/ia-topnav/src/ia-topnav.js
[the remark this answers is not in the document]
shaneriley: They're mutually exclusive

packages/ia-topnav/src/ia-topnav.js
```diff
@@ -0,0 +1,205 @@
...
+    super();
+    this.config = {};
+    this.mediaMenuOpen = false;
+    this.searchIn = '';
+    this.searchMenuAnimate = false;
+    this.searchMenuOpen = false;
+    this.signedOutMenuOpen = false;
+    this.userMenuAnimate = false;
+    this.userMenuOpen = false;
+    this.mediaSliderOpen = false;
```
jbuckner, line 43: _Suggestion_ I think all of these `*Open` and `*Animate` properties would work better as methods on the target classes ie `userMenu.open()` and `userMenu.close()`. This class is having to keep a lot of these properties in sync whenever changes occur.

To keep track of the one that's currently open, you could use a single property like `currentlyOpenMenu` and whenever the menu changes you could update that one property.

packages/ia-topnav/src/wayback-search.js
```diff
@@ -0,0 +1,61 @@
...
+  constructor() {
+    super();
+    this.config = {};
+    this.locationHandler = () => {};
+  }
+
+  handleSubmit(e) {
+    e.preventDefault();
+    const query = e.target.querySelector('#url').value;
+    this.locationHandler(`https://${this.config.waybackUrl}/web/*/${query}`);
```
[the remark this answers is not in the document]
jbuckner: When I think of a "handler" object, I tend to think of it as one that can manage a whole category of some functionality vs a single function. In this case, a good parallel would be the `window.history` handler. It manages all of the history functionality ie `forward()`, `back()`, `pushState()`, etc. 

It might just be a naming difference, but I think "handler" has a much broader/vague meaning in this context. Usually when you call a function, there's a verb involved (ie `setLocation()`, `pushState()`) vs `this.locationHandler('...')`. By just reading the name of that call, it doesn't really tell me the action that's going to be taken.

I think the second part of this is passing a function in verb form into your component/object reads kind of weird:
```
<wayback-search .setLocation=${this.setLocation}> // kind of odd here

// wayback-search.js
this.setLocation('some_url'); // reads okay here
```
vs
```
<wayback-search .locationHandler=${this.locationHandler}>

// wayback-search.js
this.locationHandler.setLocation('some_url');
```

packages/ia-topnav/README.md
```diff
@@ -0,0 +1,122 @@
...
+    --white: #fff;
+    --grey13: #222;
+    --grey20: #333;
+    --grey28: #474747;
+    --grey60: #999;
+    --grey80: #ccc;
+
+    --linkColor: #428bca;
+    --linkHoverColor: var(--white);
+    --subnavLinkColor: #aaa;
```
bfalling, line 37: QUESTION: Why were the `--linkColor` and `--subnavLinkColor` values not given color names above, so that `--linkColor` and `--subnavLinkColor` could be set to them?

bfalling: Also applies down below in `packages/ia-topnav/src/styles/ia-topnav.js`.

packages/ia-topnav/index.html
```diff
@@ -25,13 +25,22 @@
...
     </main>
     <script>
-      const topnav = document.querySelector('topnav-element');
+      const config = {
+        baseUrl: "archive.org",
+        waybackUrl: "web.archive.org",
+        screenName: "really_long_screen_name_that_may_be_truncated_on_mobile",
+        username: "shaneriley",
+        eventCategory: "MobileTopNav",
+        waybackPagesArchived: "424 billion"
```
jbuckner, line 39: Should we add the `waybackPagesArchived` to the README config example?

packages/ia-topnav/src/desktop-subnav.js
```diff
@@ -0,0 +1,44 @@
...
+        </li>
+      `
+    ));
+  }
+
+  static iconFor(label) {
+    const subnavIcons = {
+      Donate: icons.donate
+    };
+    return subnavIcons[label] ? subnavIcons[label] : html``;
```
jbuckner, line 32: How come you construct that `subnavIcons` object, then pull the label from that? Can you reference them directly from `icons.donate`?

packages/ia-topnav/src/media-button.js
```diff
@@ -0,0 +1,100 @@
...
+  static get properties() {
+    return {
+      config: { type: Object },
+      icon: { type: String },
+      href: { type: String },
+      label: { type: String },
+      mediatype: { type: String },
+      openMenu: { type: String },
+      selected: { type: Boolean },
+      followable: { type: Boolean },
```
jbuckner, line 21: What does this `followable` property do?

packages/ia-topnav/src/media-button.js
```diff
@@ -0,0 +1,100 @@
...
+    // ensure the media subnav is open on mobile if the viewport is
+    // resized, the openMenu needs to be set to 'media'.
+    if (this.openMenu !== 'media') {
+      this.dispatchEvent(new CustomEvent('menuToggled', {
+        bubbles: true,
+        composed: true,
+        detail: {
+          menuName: 'media'
+        }
+      }));
```
jbuckner, line 54: Minor, but one thing I like to do is pull out the "dispatch" calls when they get big like this. It lets you keep the "functional" flow of this method small, ie:
```
if (this.openMenu !== 'media') {
  this.dispatchMenuToggledEvent();
}
this.dispatchMediaTypeSelectedEvent();
```

packages/ia-topnav/src/media-slider.js
```diff
@@ -0,0 +1,49 @@
...
+    this.mediaSliderOpen = false;
+    this.selectedMenuOption = 'texts';
+  }
+
+  shouldUpdate() {
+    const scrollPane = this.shadowRoot.querySelector('.information-menu');
+
+    if (!scrollPane) { return true; }
+    scrollPane.scrollTop = 0;
+    return true;
```
jbuckner, line 31: Since you're returning true regardless, you could just set `scrollTop` when it exists:
```suggestion
    if (scrollPane) {
      scrollPane.scrollTop = 0;
    }
    return true;
```

packages/ia-topnav/src/nav-search.js
```diff
@@ -0,0 +1,67 @@
...
+
+class NavSearch extends TrackedElement {
+  static get styles() {
+    return navSearchCSS;
+  }
+
+  static get properties() {
+    return {
+      config: { type: Object },
+      focusSearchInput: { type: Boolean },
```
jbuckner, line 14: Is `focusSearchInput` being used?

packages/ia-topnav/src/styles/dropdown-menu.js
jbuckner: Oh I mean like what are you doing with it. Like what are you changing the top offset for?

jbuckner:
Are the tests passing for ia-topnav and just failing on the example component?

packages/ia-topnav/src/desktop-subnav.js
shaneriley: There was previously (and possibly added again in the future) an upload link that was a standalone link within the subnav, so this made it easier to determine which to render.

packages/ia-topnav/src/media-button.js
shaneriley: Whether the click/tap default is prevented. Allows for the donate link (and previously the upload link)

packages/ia-topnav/src/media-slider.js
shaneriley: Yeah, this is a `return true` that was required by ESLint that I just added in to shut it up. Previously, I only had one return statement.

packages/ia-topnav/src/nav-search.js
shaneriley: No, not anymore.

packages/ia-topnav/README.md
shaneriley: The blue color was one that was only used once and I had no idea what specific type of blue it is. The subnav one is because it's not used more than once, so I didn't see the need for setting a variable to set a variable.

shaneriley:
Closed via #430

PR #427, opened 2020-04-14
Removes 2up liner notes option

shaneriley:
**Description**

Removes 2up liner notes option from BookReader toolbar within the liner notes section of the audio player. Since this requires a newer version of BookReader than what archive.org is using, Storybook now uses a local instance of BookReader instead.

**Technical**

To serve BookReader's assets locally via Storybook, a new Node script was added to packages/ia-components/bin/copy_bookreader.js that is run as part of the `yarn run storybook` series of scripts. It copies the BookReader directory from node_modules to a new public directory. When Storybook is started, a new flag `-s` is added with a reference to the static files directory we want Storybook to serve up as root. In this instance, it's ./public.

**Testing**

Check out branch, start up Storybook as usual. Visit the theatre for audio player. The 2up control should not be present.

**Evidence**

Screen Shot 2020-04-14 at 12 28 36 PM


iisa:
getting this error:

image

PR #432, opened 2020-05-01
Sets hard pixel offset to ensure user menu offscren

shaneriley:
**Description**

Fixes user menu peeking in mobile viewports in landscape orientation.

### DIFF
--- a/packages/mobile-top-nav/src/user-menu.js
+++ b/packages/mobile-top-nav/src/user-menu.js
@@ -5,7 +5,7 @@ class UserMenu extends TrackedElement {
   static get styles() {
     return css`
       :host {
-        --topOffset: -70vh;
+        --topOffset: -1300px;
       }
       nav {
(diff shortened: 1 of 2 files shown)


packages/mobile-top-nav/src/user-menu.js
```diff
@@ -5,7 +5,7 @@ class UserMenu extends TrackedElement {
   static get styles() {
     return css`
       :host {
-        --topOffset: -70vh;
+        --topOffset: -1300px;
```
bfalling, line 8: QUESTION: Will this also work when we go to 4k screens?

shaneriley: Yes. Moving from relative to absolute units will ensure the element is in the same position no matter the viewport height. Since the user menu is shorter than this value, it will always be fully off screen.

PR #434, opened 2020-05-07
Adds hideSearch config param to disable search form and menu

shaneriley:
**Description**

Adds a `hideSearch` param on the config object to disable rendering of the nav-search and search-menu elements.

**Testing**

This will be testable with the logged in nav on https://www-shaner4.archive.org/, however to fully test you need to view the logged out state of the nav as well, which currently can only be done in this component's dev environment. To test, run `yarn start` and edit the config in index.html to remove the username and add `hideSearch: true`.

### DIFF
--- a/packages/ia-topnav/src/primary-nav.js
+++ b/packages/ia-topnav/src/primary-nav.js
@@ -85,21 +85,30 @@ class PrimaryNav extends TrackedElement {
     return this.openMenu === 'search';
   }
 
+  get searchMenu() {
+    if (this.config.hideSearch) {
+      return html``;
+    }
+    return html`
+      <button class="search-trigger" @click="${this.toggleSearchMenu}" data-event-click-tracking="${this.config.eventCategory}|NavSearchOpen">
+        ${icons.search}
+      </button>
+      <nav-search
+        .config=${this.config}
+        .open=${this.searchMenuOpen}
+        .openMenu=${this.openMenu}
+        .searchIn=${this.searchIn}
+      ></nav-search>
+    `;
+  }
+
   render() {
     const mediaMenuTabIndex = this.openMenu === 'media' ? '' : '-1';
     return html`
       <nav>
-        <a class="link-home" href="https://${this.config.baseHost}" @click=${this.trackClick} data-event-click-tracking="${this.config.eventCategory}|NavHome">${icons.iaLogo}${logoWordmark}</a>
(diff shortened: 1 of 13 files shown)


packages/ia-topnav/src/primary-nav.js
```diff
@@ -85,20 +85,29 @@ class PrimaryNav extends TrackedElement {
     return this.openMenu === 'search';
   }
 
+  get searchMenu() {
+    if (this.config.hideSearch) {
+      return html``;
+    }
+    return html`
+      <button class="search-trigger" @click="${this.toggleSearchMenu}" data-event-click-tracking="${this.config.eventCategory}|NavSearchOpen">
```
mc2, line 93: Suggestion: line length
`<button 
  class="search-trigger" 
  @click="${this.toggleSearchMenu}"
  data-event-click-tracking="${this.config.eventCategory}|NavSearchOpen"
>
${icons.search}
</button>`

packages/ia-topnav/src/signed-out-dropdown.js
```diff
@@ -5,6 +5,12 @@ class SignedOutDropdown extends DropdownMenu {
   static get styles() {
     return [DropdownMenu.styles, signedOutDropdownStyles];
   }
+
+  get menuClass() {
+    let menuClass = super.menuClass;
+    menuClass += this.config.hideSearch ? ' search-hidden' : '';
+    return menuClass;
```
mc2, line 12: Question: would this be more clean?
```
    return super.menuClass + this.config.hideSearch ? ' search-hidden' : '';
```

packages/ia-topnav/src/user-menu.js
```diff
@@ -20,6 +20,12 @@ class UserMenu extends DropdownMenu {
     this.username = '';
   }
 
+  get menuClass() {
```
mc2, line 23: Question: Any way to DRY this?

packages/ia-topnav/src/signed-out-dropdown.js
shaneriley: Yeah, it could be shortened to something like `return \`${super.menuClass}${this.config.hideSearch ? ' search-hidden' : ''}\`;` but it felt like it would be easier to quickly understand what was being returned when written in multiple statements.

packages/ia-topnav/src/user-menu.js
shaneriley: Yeah, since they're both subclassed from the same class and they're the only two doing so, I moved it to the super class method instead.

bfalling:
Code LGTM

PR #435, opened 2020-05-08
Adds production host constant for default baseHost value

shaneriley:
**Description**

On archive.org, the static markup of ia-topnav rendered via Rendertron is inserted as a child of ia-topnav. When the custom elements are later defined, each of those elements are separately initialized with an empty config object. As a result, the `baseUrl` config param is undefined and some assets no longer point to the correct location. This introduces a constant, `prodUrl`, that is used as a fallback for the primary and sub navigation items.

**Testing**

Run the app and remove `baseUrl` from index.html. The links and image sources should all point to `prodUrl`.

### DIFF
--- a/packages/ia-topnav/src/constants.js
+++ b/packages/ia-topnav/src/constants.js
@@ -1,1 +1,1 @@
-export const prodUrl = 'archive.org';
+export const prodHost = 'archive.org';
--- a/packages/ia-topnav/src/constants.js
+++ b/packages/ia-topnav/src/constants.js
@@ -1,0 +1,1 @@
+export const prodUrl = 'archive.org';
--- a/packages/ia-topnav/src/data/menus.js
+++ b/packages/ia-topnav/src/data/menus.js
@@ -1,322 +1,322 @@
-import { prodUrl } from '../constants';
+import { prodHost } from '../constants';
 
-const texts = (baseUrl = prodUrl) => ({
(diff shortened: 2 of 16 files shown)


packages/ia-topnav/src/constants.js
```diff
@@ -0,0 +1 @@
+export const prodUrl = 'archive.org';
```
bfalling, line 1: SUGGESTION: Rather than it being a complete URL, I wonder if this constant is more clearly described as the "production hostname", since it's not really a complete URL. Also, I generally prefer not using abbreviations unless absolutely necessary or unless the abbreviation is *so* standard that there is zero cognitive load in understanding it. For me, "prod" is in the gray zone; I'd probably err on the side of being explicit.

So, I wonder if `productionHostname` or `productionHost` would be clearer for this constant.

packages/ia-topnav/src/data/menus.js
```diff
@@ -1,4 +1,6 @@
-const texts = baseUrl => ({
+import { prodUrl } from '../constants';
+
+const texts = (baseUrl = prodUrl) => ({
```
bfalling, line 3: SUGGESTION: Here and throughout, again wondering if `baseHost` or `baseHostname` would be more accurate than `baseUrl`.

bfalling:
LGTM

PR #442, opened 2020-05-15
Adds location handler for searching TV captions

shaneriley:
TV caption search is handled through a query param from the TV details page rather than search.php. In the event TV is the selected search type, the `locationHandler` function should be used to redirect to the TV details page with the query as a param in the form of `?q=query`


traceypooh:
LGTM.  tested fine.  i dont understand the test, particularly, but that's OK since this is my first time reviewing anything in iaux :p

PR #443, opened 2020-05-19
Hides upload and user labels below 990px

shaneriley:
In order to provide enough horizontal space for all nav items, at < 990px viewport width the upload and username text labels should be hidden. This behavior exists on the current live nav, and provides just enough room for the signed out version of nav.

### DIFF
--- a/packages/ia-topnav/src/primary-nav.js
+++ b/packages/ia-topnav/src/primary-nav.js
@@ -114,7 +114,10 @@ class PrimaryNav extends TrackedElement {
       <nav>
         <a class="link-home" href="https://${this.config.baseHost}" @click=${this.trackClick} data-event-click-tracking="${this.config.eventCategory}|NavHome">${icons.iaLogo}${logoWordmark}</a>
         ${this.searchMenu}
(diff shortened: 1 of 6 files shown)


packages/ia-topnav/src/primary-nav.js
```diff
@@ -114,7 +114,10 @@ class PrimaryNav extends TrackedElement {
       <nav>
         <a class="link-home" href="https://${this.config.baseHost}" @click=${this.trackClick} data-event-click-tracking="${this.config.eventCategory}|NavHome">${icons.iaLogo}${logoWordmark}</a>
         ${this.searchMenu}
-        <a href="https://${this.config.baseHost}/upload/" class="upload">${icons.upload}Upload</a>
+        <a href="https://${this.config.baseHost}/upload/" class="upload">
+          ${icons.upload}
+          <span>Upload</span>
```
bfalling, line 119: QUESTION: Does putting these on their own lines introduce any undesired whitespace text nodes around the icon and text link?

shaneriley: No, in fact one of the feedback items was that there was too little space between the icon and text. This allows me to not have to add padding to the child node.

PR #444, opened 2020-05-20
Creates standalone ia-wayback-search component

shaneriley:
Extracts Wayback search component from ia-topnav to separate installable component for future use on archive.org homepage.


packages/ia-wayback-search/src/example-webcomponent.js
```diff
@@ -0,0 +1,52 @@
+import { LitElement, html, css } from 'lit-element';
+
+export default class HelloComponent extends LitElement {
```
jbuckner, line 3: Is this example component supposed to be here?

packages/ia-wayback-search/src/ia-wayback-search.js
```diff
@@ -0,0 +1,80 @@
...
+    super();
+    this.waybackPagesArchived = '';
+    this.waybackHost = 'archive.org';
+  }
+
+  handleSubmit(e) {
+    e.preventDefault();
+    const query = e.target.querySelector('#url').value;
+    this.emitWaybackSearchSubmitted(query);
+    this.locationHandler.submitCallback(`https://${this.waybackHost}/web/*/${query}`);
```
jbuckner, line 28: _Suggestion_ I would try to decouple the UI from the actual event that takes place when a search is submitted. For instance, in petabox we might want the URL redirect to happen, but in offshoot, it might cause something else like an XHR request at a different URL.

You could have this component just emit the `waybackSearchSubmitted` event like it already does, then in the consumer just listen for that event and do whatever is appropriate for that consumer (ie a `window.location` in petabox).

packages/ia-wayback-search/src/example-webcomponent.js
shaneriley: Nope. I'll remove it.

packages/ia-wayback-search/src/ia-wayback-search.js
shaneriley: That was my initial suggestion in our meeting, but you mentioned passing in an object with a method that could be mocked in tests.

jbuckner: You can still take this approach passing in the handler, but I wouldn't have the component dictate the URL that it goes to.. the handler could be more generic like
```
searchHandler.performSearch(term);
```
Then the one passed in can do anything with it. Either approach works here... you're just delegating that part of the responsibility up a level.

shaneriley: Ah, so you're wanting to pull the URL building out. That makes more sense.

packages/ia-wayback-search/src/ia-wayback-search.js
```diff
@@ -0,0 +1,78 @@
...
+import searchIcon from './icon-search';
+
+class WaybackSearch extends LitElement {
+  static get styles() {
+    return waybackSearchCSS;
+  }
+
+  static get properties() {
+    return {
+      locationHandler: { type: Object },
```
jbuckner, line 12: _Suggestion_ Since the purpose of this handler in the context of this component is to handle the query execution and not necessarily the location, I might call it a `queryHandler` and when you're calling it, use something like `this.queryHandler.performQuery(query)`. `submitCallback()` doesn't really convey the action that is happening.

jbuckner:
The build is failing on this branch.

PR #445, opened 2020-05-20
WEBDEV-3319 Radio Archive Transcript Search Fix

jbuckner:
**Description**

> Fixes an issue in the Radio Archive search where we wouldn't rebuild the transcript properly.

**Technical**

> The search engine didn't always return search results in order from start to end, but when we rebuild the transcript, we do so from start to end. This would cause the transcript rebuild to stick parts of the transcript in the wrong spot. The solution was to sort the results.

**Testing**

> For certain queries, the transcript should be correct now

### DIFF
--- a/packages/radio-player/demo/radio-player-controller.ts
+++ b/packages/radio-player/demo/radio-player-controller.ts
@@ -45,7 +45,7 @@ export default class RadioPlayerController extends LitElement {
 
   private fileName = '';
 
-  private baseUrl = 'https://58-review-radio-arch-bsdafk.archive.org';
+  private baseUrl = 'https://archive.org';
 
   private searchServicePath = '/services/radio-archive/search/service.php';
 
--- a/packages/radio-player/src/search-handler/search-handler.ts
+++ b/packages/radio-player/src/search-handler/search-handler.ts
@@ -149,18 +149,19 @@ export class SearchHandler implements SearchHandlerInterface {
     const searchRanges: Range[] = await this.searchBackend.getSearchRanges(query);
     const { mergedTranscript } = this.transcriptIndex;
 
-    // Sort the search range results from start to end
-    // The search engine may return results in any order but while processing the results here,
-    // we are going from start to finish. If the search results are out of order, it rebuilds
-    // incorrect blocks of search results.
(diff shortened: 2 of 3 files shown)


packages/radio-player/demo/radio-player-controller.ts
```diff
@@ -45,7 +45,7 @@ export default class RadioPlayerController extends LitElement {
 
   private fileName = '';
 
-  private baseUrl = 'https://58-review-radio-arch-bsdafk.archive.org';
+  private baseUrl = 'https://archive.org';
```
bfalling, line 48: QUESTION: You don't want this a relative URL?

packages/radio-player/src/search-handler/search-handler.ts
```diff
@@ -149,6 +149,12 @@ export class SearchHandler implements SearchHandlerInterface {
     const searchRanges: Range[] = await this.searchBackend.getSearchRanges(query);
     const { mergedTranscript } = this.transcriptIndex;
 
+    // Sort the search range results from start to end
+    // The search engine may return results in any order but while processing the results here,
+    // we are going from start to finish. If the search results are out of order, it rebuilds
+    // incorrect blocks of search results.
+    searchRanges.sort((a: Range, b: Range) => (a.startIndex > b.startIndex ? 1 : -1));
```
bfalling, line 156: QUESTION: Doesn't `sort` also need an equality case, returning `0`? Won't cause any endless loops/probs if not?

packages/radio-player/demo/radio-player-controller.ts
jbuckner: This is for the demo app to test searching. I meant to change it awhile ago since that review app no longer exists.

packages/radio-player/src/search-handler/search-handler.ts
jbuckner: I updated the comparison to take into account the equality case, but since we have individual word matches in our search results, we shouldn't ever get two matching start indices in our results.

This did get me thinking of the case, though, if the search engine behavior changed and we did get overlapping search results. I think we'd need to figure out how to handle that in the UI. ie. Do we merge the search result ranges, break them up into individual results, etc. There's a few cases I could see us running into.

PR #446, opened 2020-05-21
IA icons component library

shaneriley:
Creates a framework for defining SVG icons for use in IAUX. The initial package has icons used in ia-topnav with one icon used in the BookReader chrome redesign for the purpose of illustrating the method of defining and changing the stroke property.


packages/ia-icons/src/ia-icons.js
```diff
@@ -0,0 +1,130 @@
...
+import upload from './icons/upload';
+import user from './icons/user';
+import video from './icons/video';
+import web from './icons/web';
+
+class Icon extends LitElement {
+  static get properties() {
+    return {
+      fill: { type: String },
+      stroke: { type: String },
```
jbuckner, line 21: Can these be customized as CSS variables instead of properties?

packages/ia-icons/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Usage
+
+```js
+import { IconClose } from './src/ia-icons';
+
+customElements.define('icon-close', IconClose);
+```
+
+```html
+<icon-close></icon-close>
```
jbuckner, line 12: Is the intended use to define each icon individually like this?

I think it would help icon discoverability and consumption if it were packaged under a single component with an enumerated list of options. ie:
```
<ia-icon icon="close"></ia-icon>
```
I'm not sure if there's a good way to do an enum of the icon options with JS, but it'd be great to have some reference of what icons are available without having to open up the readme.

packages/ia-icons/src/ia-icons.js
shaneriley: You know, when I first implemented these in ia-topnav, I couldn't use CSS variables because they were simply lit-html templates, but now that they're full-on encapsulated components I should be able to go that route. Let me give it a try.

packages/ia-icons/README.md
shaneriley: I was thinking that there would end up being so many icons that even if you only wanted one icon in your project, you'd be bringing the rest along with it.

As far as reference is concerned, if you run the server the index will render a matrix of all of the icons defined with a random fill and stroke color.
Screen Shot 2020-05-21 at 3 30 39 PM

jbuckner: Regarding only bringing a single icon in, it looks like they all get imported in that top-level `Icon` class so you're already importing them all.

Hmm it'd be nice to not have to run the project to see the reference. We'll probably be using this library while working on other things so spinning this up for reference.

shaneriley: I was thinking https://archive.org/help/coding.php would have the icons rendered once we've included it in Petabox.

packages/ia-icons/src/ia-icons.js
shaneriley: After looking at this for a bit, I can't figure out how we'd allow for a different fill and stroke color for each usage of the icon without passing them in as props. If there's multiple instances of the magnifying glass icon, for example, and each are a different fill color, they'd need a scoped CSS declaration for each that would set the fill color variable. I was trying to mimic the method of adding fill and stroke colors that SVG uses instead.

This might be something that the group should decide on before I go further with it.

jbuckner: Does it work if you use more specific selectors like this?
```
<style>
#mag1 {
  --fillColor: blue;
}
#mag2 {
  --fillColor: green;
}
</style>

<ia-icon icon="magifying" id="#mag1"></ia-icon>
<ia-icon icon="magifying" id="#mag2"></ia-icon>
```

packages/ia-icons/README.md
jbuckner, line 12: Make sure to update the readme for the new syntax.

shaneriley: :t_hanks:

packages/ia-icons/src/ia-icons.js
shaneriley: Updated to use CSS variables. I wasn't thinking about how the cascade would allow re-declaration of the same CSS variables within a tighter scope, allowing each icon to have a different color. Also added an example of having an icon component with a default color.

packages/ia-icons/README.md
```diff
@@ -0,0 +1,36 @@
+# ia-icons: SVG icon library for use in archive.org components
+
+## Usage
+
+```js
+import { IAIcon } from './src/ia-icon';
+```
+
+```html
+<ia-icon icon="close" style="--iconFillColor: #0066cc; --iconStrokeColor: #ffffff;"></ia-icon>
```
jbuckner, line 10: Oh wow that's a cool use of CSS variables!

PR #452, opened 2020-05-26
Update Radio Player to 0.0.1 and Add Tests

jbuckner:
Bump Radio Player to 0.0.1.

The CodeCov report is reporting some odd results so I had to add some unit tests too.


jbuckner:
Replaced by https://github.com/internetarchive/iaux/pull/451

PR #457, opened 2020-06-03
Wayback subnav

shaneriley:
Creates the Wayback subnav based on designs in https://webarchive.jira.com/browse/WEBDEV-3434


packages/ia-topnav/src/lib/query-handler.js
```diff
@@ -0,0 +1,7 @@
+/* istanbul ignore file */
+
+export default {
+  performQuery(query) {
+    window.location = `https://web.archive.org/web/*/${query}`;
```
bfalling, line 5: QUESTION: Does `query` need to be URL-encoded?

shaneriley: Yeah, it should. Good catch. I've run it through `encodeURIComponent`

PR #460, opened 2020-06-15
Uses fixed measurement position for media menu on mobile

shaneriley:
<media-menu> previously used vh units to position off screen for mobile breakpoints. When the viewport height was too short, however, the bottom edge would still be visible. This sets a fixed pixel value that slightly exceeds the height of the media menu to ensure it remains completely off screen when closed.

See animation, which colors the media menu red for demonstration.

media-menu


jbuckner:
Does this still work okay if you change the browser font size?

jbuckner:
Code looks good, just one question.

PR #469, opened 2020-06-23
Fixes text of software nav item

shaneriley:
Fixes (WEBDEV-3550)[https://webarchive.jira.com/browse/WEBDEV-3550]

### DIFF
--- a/packages/ia-topnav/src/data/menus.js
+++ b/packages/ia-topnav/src/data/menus.js
@@ -210,7 +210,7 @@ const software = (baseHost = prodHost) => ({
     title: 'All software',
     url: `https://${baseHost}/details/software`,
   }, {
-    title: 'latest This Just In',
+    title: 'This Just In',
     url: `https://${baseHost}/search.php?query=mediatype:software&sort=-publicdate`,
   }, {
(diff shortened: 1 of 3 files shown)


packages/ia-topnav/src/data/menus.js
```diff
@@ -210,7 +210,7 @@ const software = (baseHost = prodHost) => ({
     title: 'All software',
```
iisa, line 210: Q: we had a plan to move this output and feed the data to the web-component. is that in the pipeline?

traceypooh:
LGTM

packages/ia-topnav/src/data/menus.js
shaneriley: The plan was to have ia-topnav used as nav for all users, allowing me to remove mobile-top-nav and the nav building in Nav.inc. At that point, Nav.inc would be updated to simply send its subnav arrays to ia-topnav and it would blindly render whatever nav items you fed it.

iisa: ah, cool. so that will happen once mobile nav is gone. dope!

PR #472, opened 2020-06-25
Receives menu objects from external data

shaneriley:
Moves the menu arrays and objects out of ia-topnav and instead accepts a prop `menus` that internally sends individual menus to the concerned components. This will be necessary for v1, as I will be removing the current static nav from Nav.inc and passing the generated menu objects to ia-topnav on render.

### DIFF
--- a/packages/ia-topnav/src/data/petabox_signedout_menus.js
+++ b/packages/ia-topnav/src/data/petabox_signedout_menus.js
@@ -1,0 +1,509 @@
+/* eslint-disable */
+export default {
+  "audio": {
+    "heading": "Internet Archive Audio",
+    "iconLinks": [
+      {
+        "icon": "https://www-shaner4.archive.org/services/img/etree",
+        "title": "Live Music Archive",
+        "url": "https://www-shaner4.archive.org/details/etree"
+      },
+      {
+        "icon": "https://www-shaner4.archive.org/services/img/librivoxaudio",
+        "title": "Librivox Free Audio",
+        "url": "https://www-shaner4.archive.org/details/librivoxaudio"
+      }
+    ],
+    "featuredLinks": [
+      {
+        "title": "All audio",
+        "url": "https://www-shaner4.archive.org/details/audio"
+      },
+      {
+        "title": "This Just In",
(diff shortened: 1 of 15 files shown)


packages/ia-topnav/src/data/petabox_signedout_menus.js
```diff
@@ -0,0 +1,509 @@
+/* eslint-disable */
+export default {
+  "audio": {
+    "heading": "Internet Archive Audio",
+    "iconLinks": [
+      {
+        "icon": "https://www-shaner4.archive.org/services/img/etree",
+        "title": "Live Music Archive",
+        "url": "https://www-shaner4.archive.org/details/etree"
```
bfalling, line 9: BLOCKER: All these links are using the dev host.

packages/ia-topnav/test/media-slider.test.js
```diff
@@ -6,7 +6,8 @@ import { texts } from '../src/data/menus';
 const component = (mediaSliderOpen, selectedMenuOption) => (
   html`<media-slider
           ?mediaSliderOpen="${mediaSliderOpen}"
-          selectedMenuOption="${selectedMenuOption}"></media-slider>`
+          selectedMenuOption="${selectedMenuOption}"
```
bfalling, line 9: QUESTION: Is there a reason this parameter is treated as an attribute and not a property, e.g. `.selectedMenuOption`? Seems like in most other places in this code, it's treated as a JS property.

packages/ia-topnav/src/data/petabox_signedout_menus.js
shaneriley: 1. This is only here for testing purposes to provide a way to test Petabox data without having to base64 encode it
2. The feature this PR completes is switching from always using the static menu data in the menus.js file in this directory to always requiring the menu data be provided to it on instantiation. This means that no matter what's in this test data, it won't make it to production because it's not being imported anywhere in the module files.

packages/ia-topnav/test/media-slider.test.js
shaneriley: No, there's no reason. It was simply omitted. It will behave the same, however, since the default attribute converter is being used.

PR #484, opened 2020-08-05
Adds base64 decoding converter to config property

shaneriley:
To make it easier to transport the data from Petabox to the nav, base64 encoding has been added on the Petabox side. This adds the same type of property converter to config that is already present on the menus.


bfalling:
LGTM

PR #488, opened 2020-09-08
Fix Tests by Setting Test Concurrency

jbuckner:
**Description**

> Test if setting the test concurrency fixes the unit test failures.

**Technical**

> The reason the tests were failing is because `lerna` tries to run the tests in parallel, but if you have too many test runners running all at once, they conflict with one another. Setting the concurrency to 1 slows it down, but the tests pass.

**Testing**

> No testing needed. The unit tests should pass.

**Evidence**

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,10 @@
     "packages/*"
   ],
   "version": "independent",
-  "npmClient": "yarn"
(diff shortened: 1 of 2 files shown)


bfalling (approved):
Seems okay.

PR #499, opened 2020-10-05
Zendesk help widget

nsharma123:
**Description**

> As per request in https://webarchive.jira.com/browse/WEBDEV-3758, adding a Zendesk custom help button on archive.org site 

**Testing**

> Visit on https://webdev-3758.archive.org/ to see help button on pages.


packages/ia-zendesk-help-widget/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "@open-wc/testing-karma": "^3.4.0",
+    "@open-wc/testing-karma-bs": "^1.0.0",
+    "@storybook/react": "^5.1.11",
+    "es-dev-server": "^1.42.0",
+    "eslint": "^6.1.0",
+    "eslint-plugin-lit": "^0.5.0",
+    "gulp": "^4.0.2",
+    "husky": "^4.2.3",
+    "lint-staged": "^8.0.0",
+    "sinon": "^9.0.0",
```
shaneriley, line 48: `sinon` is never used, so it can be dropped.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,75 @@
...
+
+  constructor() {
+    super();
+    this.widgetSrc = '';
+    this.widgetkey = '';
+  }
+
+  initiateZenDesk() {
+    // Hide custom button
+    this.style.opacity = 0;
```
shaneriley, line 25: All styles should be housed within the `styles` template that is returned from the static method and a class should be used to toggle opacity instead. Typically, you would do this by updating a tracked property on the component, like `visible`. Then you can have the default style set to `opacity: 0` and the visible class override it.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,75 @@
...
+
+    // prevent further if third party script already loaded
+    if (this.abortIfScriptAlreadyLoaded()) return;
```
shaneriley, line 28: This will never return early because the return value of this method is always `undefined`.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,75 @@
...
+          // Found it! Click on it and hope the window opens!
+          button.click();
+          clearInterval(timeoutTimer);
+          return;
+        }
+      }
+
+      // Try again later
+      countdownTicksLeft--;
+      if (countdownTicksLeft <= 0) clearInterval(timeoutTimer);
```
shaneriley, line 49: We may also want to remove the component entirely from the DOM if we reach this point, just to be certain users navigating via keyboard or screen reader don't find the help button and interact with it.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,75 @@
...
+
+  injectScript() {
+    const script = document.createElement('script');
+    script.setAttribute('id', 'ze-snippet');
+    script.setAttribute('src', this.widgetSrc + this.widgetkey);
+    document.head.appendChild(script);    
+  }
+
+  abortIfScriptAlreadyLoaded() {
+    if (document.querySelector(`script[src="${this.widgetSrc + this.widgetkey}"]`)) return;
```
shaneriley, line 61: Update to return ```!!document.querySelector(`script[src="${this.widgetSrc + this.widgetkey}"]`)```

packages/ia-zendesk-help-widget/src/styles/zendesk-widget.js
```diff
@@ -0,0 +1,43 @@
...
+export default css`
+  .help-widget {
+    padding: 13px 22px;
+    margin: 14px 20px;
+    position: fixed;
+    bottom: 0px;
+    z-index: 999998;
+    transition-duration: 250ms;
+    right: 0px;
+    background: #194880;
```
shaneriley, line 12: **Suggestion**: create CSS variables for background and text color to allow overriding styles based on context.

packages/ia-zendesk-help-widget/src/icon-question-mark.js
```diff
@@ -0,0 +1,14 @@
+import { html } from 'lit-element';
+
+export default html`
+  <svg
+    height="20"
+    viewBox="0 0 20 20"
+    width="20"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <g fill="#fff" fill-rule="nonzero">
```
shaneriley, line 10: The `fill` attribute should be replaced with a class of `fill-color` to adhere to our icon standard of setting the fill and stroke colors from outside of the markup. The CSS variables `--iconFillColor` and `--iconStrokeColor` should be set and used to provide the values for fill and stroke properties in the CSS for the component.

packages/ia-zendesk-help-widget/test/zendesk-help-widget.test.js
```diff
@@ -0,0 +1,26 @@
...
+  html,
+  fixture,
+  expect,
+  oneEvent
+} from '@open-wc/testing';
+import '../src/ia-zendesk-help-widget';
+
+const component = (properties = {
+  widgetSrc: 'https://static.zdassets.com/ekr/snippet.js?key=',
+  widgetKey: '685f6dc4-48c5-411f-8463-cc6dd50abe2d',
```
shaneriley, line 11: Odd. The `widgetKey` property is camel case here, but is not in the component definition. How are these tests passing?

shaneriley (changes requested):

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,75 @@
...
+
+class ZenDeskHelpWidget extends LitElement {
+  static get styles() {
+    return zendeskWidgetCSS;
+  }
+
+  static get properties() {
+    return {
+      widgetSrc: { type: String },
+      widgetkey: { type: String },
```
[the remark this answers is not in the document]
nsharma123: My mistake. will change this.

jbuckner, line 13: _Suggestion_ Since you're just merging the key and the src into a single URL, I would just use a single parameter.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,75 @@
...
+
+  initiateZenDesk() {
+    // Hide custom button
+    this.style.opacity = 0;
+
+    // prevent further if third party script already loaded
+    if (this.abortIfScriptAlreadyLoaded()) return;
+
+    // Inject help widget
+    this.injectScript();
```
jbuckner, line 31: _Suggestion_ We have a javascript lazy loader service that allows you await the script loading like a promise so you can wait for it to load, then start your check for the button. ie:

Install it like this:
```
npm install @internetarchive/lazy-loader-service
```
then
```
import { LazyLoaderService } from '@internetarchive/lazy-loader-service';

async initiateZenDesk() {  <--- make this async
  const lazyLoaderService = new LazyLoaderService();
  await lazyLoaderService.loadScript({ src: this.widgetSrc });

  // the script is loaded, check for the button
}
```

It also won't double-load a script if it's already loaded so you won't have to check if the script has already been inserted.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,75 @@
...
+    document.head.appendChild(script);    
+  }
+
+  abortIfScriptAlreadyLoaded() {
+    if (document.querySelector(`script[src="${this.widgetSrc + this.widgetkey}"]`)) return;
+  }
+
+  render() {
+    return html`
+      <div
```
jbuckner, line 66: Since this is a button, I think it'd make more sense semantically to use a `button` tag.

packages/ia-zendesk-help-widget/src/styles/zendesk-widget.js
```diff
@@ -0,0 +1,43 @@
+import { css } from 'lit-element';
+
+export default css`
```
jbuckner, line 3: _Suggestion_ You can put your style directly inside your WebComponent so you don't have to jump between files to see what your component is doing.

packages/ia-zendesk-help-widget/src/styles/zendesk-widget.js
```diff
@@ -0,0 +1,43 @@
...
+    cursor: pointer;
+    vertical-align: middle;
+  }
+
+  .help-widget svg {
+    vertical-align: middle;
+    margin-right: 3px;
+  }
+
+  @media (max-width: 767px) {
```
jbuckner, line 30: In general, I would try to avoid setting state inside a component that is dependent on something happening outside of the component, in this case, the window being resized.

Since it looks like you're hiding the "Help" text when the window gets smaller, I would treat this as 2 different modes, minimal and full, and make that a property on the webcomponent itself, ie:
```
buttonMode: { type: String } // with values of "minimal" or "full"
```
then set a class on your component based on the mode, ie:
```
<div class="${this.buttonMode}">...</div>
```
set up your CSS:
```
css`
  .minimal .help-text {
    .display: none;
  }
  .minimal .help-width {
    padding: 13px 13px;
  }
```
Then let your consumer set the mode. 
```
<ia-zendesk-help-widget buttonMode="minimal"></ia-zendesk-help-widget>
```

Since it can change based on the window size, you can do a `window.onresize()` to change the property when it reaches the threshold.

packages/ia-zendesk-help-widget/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@internetarchive/ia-zendesk-help-widget",
+  "version": "0.1.1-alpha11",
```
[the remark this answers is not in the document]
nsharma123: Yeah I will do that once we get final code review.

packages/ia-zendesk-help-widget/src/styles/zendesk-widget.js
nsharma123: I just wanted to make in its own file so that things can be managed for future enhancements.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
nsharma123: Thanks for catching this!

packages/ia-zendesk-help-widget/src/styles/zendesk-widget.js
jbuckner: How does this help manage future enhancements?

packages/ia-zendesk-help-widget/test/zendesk-help-widget.test.js
nsharma123: Changed these

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
nsharma123: Thanks

packages/ia-zendesk-help-widget/src/styles/zendesk-widget.js
nsharma123: I think that will increase execution time to traverse things with window.onresize event and all. we generally can use media queries as well in CSS components so I am skipping this.

jbuckner: Even though you _can_ use the media query inside a component, you are ignoring component layout encapsulation by doing so. The component, particularly one that isn't intended to take up the whole width, should only be concerned with its own layout, given its container size (or mode as the case may be), not with the entire page size. There are APIs like the [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) that help with this.

packages/ia-zendesk-help-widget/src/styles/zendesk-widget.js
nsharma123: The main component file looks optimised because we pulled CSS in separate file. and later if we want to make more changes and want to add more CSS, that will consume more space in main file.

jbuckner (changes requested):

packages/ia-zendesk-help-widget/src/styles/zendesk-widget.js
nsharma123: If I understood you correctly, we want to make changes something like 
`static get styles() {
    return css`
      :host {
        --buttonBlue: #194880;
        --white: #fff;

        --iconFillColor: var(--white);
        --linkColor: var(--white);
      }`

nsharma123: I will change something like this so that execute won't switch over files to get styling for this.

packages/ia-zendesk-help-widget/src/styles/zendesk-widget.js
```diff
@@ -0,0 +1,62 @@
...
+
+    --iconFillColor: var(--white);
+    --linkColor: var(--white);
+  }
+
+  .help-widget {
+    width: 108px;
+    height: 46px;
+    margin: 14px 20px;
+    position: fixed;
```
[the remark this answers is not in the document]
nsharma123: Since help button rendering on DOM independently at right-bottom corner, not sure how without fixed position we can make this happen?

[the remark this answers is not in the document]
jbuckner: You can let the consumer set the positioning. You can still use `fixed`, but it should be the consumer (petabox) that sets that, not the component itself.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,141 @@
...
+      if (countdownTicksLeft <= 0) clearInterval(timeoutTimer);
+    }, 250);
+  }
+
+  get buttonVisibilityState() {
+    return !this.buttonVisible ? 'hidden' : '';
+  }
+
+  get getActiveIcon() {
+    return !this.isLoading ? loadingIcon : questionIcon
```
jbuckner, line 126: Is this logic correct? "If this is not loading, use the loading icon, otherwise use the question icon"

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,141 @@
+import { html, css, LitElement } from 'lit-element';
+import loadingIcon from './assets/icon-loading';
+import questionIcon from './assets/icon-question-mark';
+import { LazyLoaderService } from '@internetarchive/lazy-loader-service';
+
+class ZenDeskHelpWidget extends LitElement {
+  static get styles() {
```
jbuckner, line 7: I find it useful to group the `styles` method and the `render` method since it's those two things that are interacting the most. Also, since the `styles` section is so much bigger than the rest, I like putting it at the bottom to not obscure all the business logic, ie:
 
```
class ZenDeskHelpWidget extends LitElement {
  ... business logic ...
  render() { }
  styles() { }
}
```

shaneriley: Putting the styles in a separate file also helps to not obscure all of the business logic, which is the primary reason why I separate them.

jbuckner: That also misses the first point:
> I find it useful to group the styles method and the render method since it's those two things that are interacting the most.

nsharma123: Instead of putting them in same file, we should consider styles in separate files as it was earlier since CSS code is bigger part.

packages/ia-zendesk-help-widget/src/styles/zendesk-widget.js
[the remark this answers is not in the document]
nsharma123: Since this component is for render overlayed help button and most of the help icon renders at bottom-right corner so it doesn't make sense to use another property for just positioning. so I am keeping this as it is.

jbuckner:
Could you take a look at the unit test coverage for this package? It looks like the tests are failing because the coverage is low. If you run `npm run test` then take a look at `coverage/index.html`, you can see what lines are missing code coverage.
---
When QA'ing, I noticed there is still a delay the first time after hiding the help button before the popup appears:
zendesk-delay

packages/ia-zendesk-help-widget/test/zendesk-help-widget.test.js
```diff
@@ -0,0 +1,89 @@
+import {
+  html,
+  fixture,
+  expect
+} from '@open-wc/testing';
+import '../src/ia-zendesk-help-widget';
+
+const component = (properties = {
+  widgetSrc: 'https://static.zdassets.com/ekr/snippet.js?key=685f6dc4-48c5-411f-8463-cc6dd50abe2d',
```
jbuckner, line 9: Is this making a request to the live zen desk site during the unit test? If so, I would recommend replacing it with a local mock javascript file that simulates the part of the zen desk widget that you are interacting with. We don't really want to be making network requests during unit testing.

If you add the mock widget to this directory, you can set your `widgetSrc` to `/base/test/mock-zen-desk.js` and it will find it. You can see some examples in [the unit tests for the lazy loader service](https://github.com/internetarchive/iaux-lazy-loader-service/blob/master/test/lazy-loader-service.test.ts#L5-L13).

packages/ia-zendesk-help-widget/README.md
```diff
@@ -0,0 +1,40 @@
+# Zendesk Help Widget Component
+
+## Usage
+
+```html
+<ia-zendesk-help-widget
+  widgetSrc="https://static.zdassets.com/ekr/snippet.js?key="
+  buttonVisible="true"
```
jbuckner, line 8: If the button is not visible, couldn't you just not add it to the DOM instead of setting a property on it?

packages/ia-zendesk-help-widget/test/mock-zen-desk.js
```diff
@@ -0,0 +1,7 @@
+const iframe = document.createElement('iframe');
+iframe.setAttribute('id', 'launcher');
+document.body.appendChild(iframe);
+const element = '<body><button>help</button></body>';
+iframe.contentWindow.document.open();
+iframe.contentWindow.document.write(element);
+iframe.contentWindow.document.close();
```
jbuckner, line 7: Nice 👏

packages/ia-zendesk-help-widget/test/zendesk-help-widget.test.js
```diff
@@ -0,0 +1,80 @@
+import {
+  html,
+  fixture,
+  expect
+} from '@open-wc/testing';
+import '../src/ia-zendesk-help-widget';
+
+const testwidgetSrc = '/base/test/mock-zen-desk.js';
```
jbuckner, line 8: Nice improvement on your tests!

packages/ia-zendesk-help-widget/README.md
jbuckner, line 8: If the button is not visible, couldn't you just not add it to the DOM?

nsharma123: Good suggestion, lets just not add whole stuff at petabase side.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,148 @@
...
+    return {
+      widgetSrc: { type: String },
+    };
+  }
+
+  constructor() {
+    super();
+    this.widgetSrc = '';
+    this.buttonVisible = true;
+    this.isLoading = false;
```
jbuckner, line 17: I'm not sure if this was intentional, but if you want `buttonVisible` and `isLoading` to be reactive, add them to the `static get properties() {}`. Right now if you change either of them, the UI won't update.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  async initiateZenDesk() {
+    this.isLoading = true;
+
+    // force UI update to render loader icon after button click
+    this.requestUpdate();
```
jbuckner, line 24: You shouldn't have to force an update like this. If you make `isLoading` reactive, it will update automatically.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,146 @@
...
+  get getActiveIcon() {
+    return this.isLoading ? loaderIcon : questionIcon;
+  }
+
+  render() {
+    return html`
+      <button
+        class="help-widget ${this.buttonVisibilityState}"
+        @click=${this.initiateZenDesk}
+        data-event-click-tracking="ZenDesk|InitialHelpButton"
```
jbuckner, line 71: Is this event click tracking doing anything? I don't see anything consuming it.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,146 @@
...
+          return;
+        }
+      }
+
+      // Try again later
+      countdownTicksLeft -= 1;
+      /* istanbul ignore if */
+      if (countdownTicksLeft <= 0) clearInterval(timeoutTimer);
+    }, 250);
+  }
```
jbuckner, line 56: On the petabox side, I see
```
helpButton.addEventListener('ZendeskHelpButton', () => {})
```
But I don't see you emitting an event when the button gets clicked. Are you expecting to be tracking an event?

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.js
```diff
@@ -0,0 +1,151 @@
...
+
+  emitZendeskHelpButtonClicked() {
+    this.dispatchEvent(new CustomEvent('zendeskHelpButtonClicked'));
```
jbuckner, line 61: Not a blocker, but if you're just emitting an event without any parameters, you can do `new Event('foo')` instead of `CustomEvent('foo')`

packages/ia-zendesk-help-widget/test/zendesk-help-widget.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+    el.widgetSrc = testwidgetSrc;
+    el.buttonVisible = false;
+
+    await el.updateComplete;
+
+    expect(el.buttonVisibilityState).to.equal('hidden');
+  });
+
+  it('terminate before completing interval', (done) => {
```
jbuckner, line 63: This test isn't testing your element.

packages/ia-zendesk-help-widget/test/zendesk-help-widget.test.js
```diff
@@ -0,0 +1,70 @@
...
+    if (counter >= 5) {
+      testableVariable = true;
+      clearInterval(interval);
+    }
+  }, 200);
+
+  return interval;
+}
+
+describe('<zendesk-help-widget>', () => {
```
jbuckner, line 34: Not a blocker, but one spot I think could use unit testing is the UI change when the media query changes.

jbuckner (approved):

PR #500, opened 2020-10-14
Extracts base host to prop

shaneriley:
**Description**

In order to allow the change of the base host for most links in the topnav, a property needs to be added to the ia-topnav component. This is also passed in to the ia-wayback-search component, which receives the same property addition. A function is added to lib to check whether the URL being used is fully qualified or not, and if not, uses the baseHost property to build a full URL.

**Testing**

Visit https://www-shaner5.archive.org/. All URLs that would normally be root relative use the www-shaner5 subdomain. If you then change the property with `document.querySelector('ia-topnav').baseHost = 'archive.onion'`, those same URLs should update to use the new baseHost.

### DIFF
--- a/packages/ia-topnav/README.md
+++ b/packages/ia-topnav/README.md
@@ -78,14 +78,13 @@ export default IATopNav;
   }
 </style>
 
-<ia-topnav></ia-topnav>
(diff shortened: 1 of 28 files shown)


jbuckner (approved):
Nice! 👍 Setting the `baseHost` with JS to update the whole menu is pretty slick!

PR #501, opened 2020-11-04
Fix TopNav Signed Out Bug

jbuckner:
**Description**

> Fixes a bug in mobile view where the Sign Up and Log In buttons went to an incorrect URL

**Technical**

> We weren't passing the baseHost into the signed out dropdown


iisa (approved):
LGTM thank you!

PR #510, opened 2021-01-19
Added expand title on topnav media buttons

nsharma123:
**Description**
https://webarchive.jira.com/browse/WEBDEV-3856

> Added title attribute on media button in topnav so whenever user mouse over, it will say "Expand texts/video/audio menu".


jbuckner (approved):
Looks good 👍

PR #512, opened 2021-02-12
Ia components 0.0.6

iisa:


packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify.stories.js
```diff
@@ -92,7 +92,7 @@ class DataHydrator extends Component {
     // using bootstrap v3 styling for container for pretty UI
     return (
       <div>
-        <section>
+      <section>
```
bfalling, line 95: SUGGESTION: Seems like this should still be indented.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
```diff
@@ -49,11 +49,19 @@ export default class BookReaderWrapper extends Component {
...
+      /**
+       * Needed bypass to generate Image URL with scale factor.
+       * We must eliminate sooner than later to allow BookReader full image fetching control
+       * @param {Number} index - page index
+       * @param {Number} reduce - image size scale factor
+       * @param {Number} rotate - degrees of rotation
+       */
       getPageURI: (index, reduce = 1, rotate = 0) => {
-        let uri = originalGetPageURI.call(br, index, reduce, rotate);
+        const brReduce = this.bookreader.reduce || reduce;
```
bfalling, line 61: QUESTION: So just checking: `this.bookreader.reduce` will now take precedence over any passed-in `reduce`, right?

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
```diff
@@ -73,6 +81,7 @@ export default class BookReaderWrapper extends Component {
     } else {
       const br = new BookReader(fullOptions);
       window.br = br;
+      this.bookreader = window.br;
```
bfalling, line 84: QUESTION: Why is `br` effectively being stored in both `window.br` and `this.bookreader`?

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
```diff
@@ -38,7 +39,6 @@ export default class BookReaderWrapper extends Component {
     const originalGetPageURI = BookReader.prototype.getPageURI;
     const defaultOptions = {
       el: `#${this.BookReaderRef.current.id}`,
-      mobileNavFullscreenOnly: true,
```
bfalling, line 41: QUESTION: Why was this removed?

bfalling (approved):
Some questions. No blockers.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
iisa: yes.  this URL string method bypass is NOT the most optimum. in fact we should not even have it at all - but at the moment, we must hold.

We never want the URL to be requested without its image scale, unless you are extremely zoomed, in - that case we handle well enough.

BR 1up mode - sets reducer & reduction factors that allow zooming - and we always want to prefer the calculated image scale at this level. like we did for 2up mode.

packages/ia-components/sandbox/theatres/with-youtube-spotify/audio-with-youtube-spotify.stories.js
iisa: yes, unsure why this is sticky. will fix.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
iisa: window.br is so globally used, we must keep it until we can truly remove the global dependency.

packages/ia-components/sandbox/bookreader-component/bookreader-wrapper-main.jsx
iisa: thanks for catching. i've re-added.
did not mean to remove until we move into item-navigator world

PR #515, opened 2021-03-08
Update Topnav Base Host handling

jbuckner:
**Description**

> Previously, the top nav was writing out an `https://` prefix to all of the URLs so we couldn't make the links relative if we wanted to.

**Technical**

> This changes the base URL so the consumer should now pass in the protocol as well

**Testing**

> Checkout the repo, go to `packages/ia-topnav`, run `yarn install`, run `yarn start` and you should see a demo of it. You can click the "Change Base Host" button in the demo and it will change the URLs to use the new base host


iisa (approved):
LGTM

PR #531, opened 2021-03-18
[bookreader-component] getPageURI fixes

cdrini:
**Description**

- getPageURI should use argument instead of internal bookreader property
- getPageURI should only use powers of 2 for reduce

**Technical**

Blocker for https://github.com/internetarchive/bookreader/pull/646

**Testing**

Available for testing with https://github.com/internetarchive/bookreader/pull/646 at e.g. https://ia-petabox-drini.archive.org/details/lp_chant-of-the-jungle_august-colon . Things to note:

1) Zooming in loads higher res copies.
2) Urls for images only are powers of 2.
3) As you zoom in, fewer requests need to be made since using powers of 2 now.

**Evidence**


iisa (approved):
LGTM

PR #552, opened 2021-04-15
[Transcript View] Remove leading 0 in the duration formatter

jbuckner:
**Description**

> Removes the leading 0 in the duration formatter. ie: 02:13:43 will now be 2:13:43 and 02:43 will be 2:43

**Testing**

> The unit tests cover the new formats

**Evidence**

> 
Screen Shot 2021-04-15 at 4 01 41 PM

Screen Shot 2021-04-15 at 4 01 46 PM


iisa (approved):
looks good!
parsing time is always so...fun 😅

Issue #528, opened 2021-03-16
Selected quick search entry doesn't show up in input field after search is cleared


Swapnil-2001:
**Description**

> What problem are we solving? What does the experience look like today? What are the symptoms? How can one reproduce the problem?
In the **expandable search bar**, where we have a quick search functionality, selecting a quick search item does not make it show up in the input field, **after** the `clear search` event has been triggered once. As long as the `clear search` is not triggered, it works fine.

**Evidence / Screenshot (if possible)**

https://user-images.githubusercontent.com/53232360/111375983-ea723f80-86c4-11eb-91f5-811e79cf9d32.mp4

> Include evidence of the problem or provide a screenshot.

**Expectation**

> What should be happening? What will it look like / how will it behave when fixed? 

The selected quick search item should show up in the input field as expected.

**Context**

> What operating system? What browser type and version?

Windows 10, Google Chrome Version 89.0.4389.82

**Proposal & Constraints**

> What is the proposed solution / implementation? Is there a precedent of this approach succeeding elsewhere? 

The issue might be with the `clear search` method's implementation.

**Stakeholders**

> @-tag any stakeholders of this bug

PR #553, opened 2021-04-18
fix expandable search bar issue w/ the dropdown options (quick search)

hotzjacobb:
**Description**

This PR addresses issue #528. The issue was actually broader than described. The expandable search bar was actually unable to use the dropdown options after any user input keystrokes. This PR fixes that and it now works as intended.

**Technical**

The existing code didn't work because it tried to modify a LitElement component from outside the component. LitElement, like similar frameworks, updates the DOM based on changes to state inside components. Also, the previous implementation placed a non-documented obligation on the consumer to know that they must write this supporting code.

This implementation fixes this by changing state from within the component.

I also would like to note that I discovered another pre-existing problem while fixing this, but that goes beyond the scope of this PR. Both `this.searchTerm` and `this.searchInput.value` hold overlapping state information for text in the search bar. I tried to see if there was a simple fix by looking at LitElement's documentation, but it seemed like it would take more digging or someone with more experience with that framework. A simple solution might be converting this to React. I can file an issue if need be.

**Testing**

See if the expandable search bar correctly takes new input after from the drop-down menu/the expandable search bar works in general.

**Evidence**

https://user-images.githubusercontent.com/34525444/115131511-85ed1d80-9fad-11eb-8a2e-593d73874ad0.mov

### DIFF
--- a/packages/expandable-search-bar/src/expandable-search-bar.ts
+++ b/packages/expandable-search-bar/src/expandable-search-bar.ts
@@ -29,6 +29,8 @@ export default class ExpandableSearchBar extends LitElement {
 
   @property({ type: Boolean }) showsDisclosure = false;
 
+  // TODO: searchTerm should be single source of state;
+  //       searchInput.value should not also affect state
   @property({ type: String }) searchTerm = '';
 
   @property({ type: Array }) quickSearches: QuickSearchEntry[] = [];
@@ -165,6 +167,11 @@ export default class ExpandableSearchBar extends LitElement {
    * @memberof ExpandableSearchBar
    */
   private quickSearchSelected(e: CustomEvent): void {
+    this.searchTerm = e.detail.searchEntry.displayText;
+    /* istanbul ignore else */
+    if (this.searchInput) {
+      this.searchInput.value = e.detail.searchEntry.displayText;
+    }
     const event = new CustomEvent('quickSearchSelected', {
       detail: { quickSearchEntry: e.detail.searchEntry },
     });
--- a/packages/expandable-search-bar/src/expandable-search-bar.ts
+++ b/packages/expandable-search-bar/src/expandable-search-bar.ts
@@ -86,7 +86,7 @@ export default class ExpandableSearchBar extends LitElement {
    *
    * @memberof ExpandableSearchBar
    */
-  updated(props: PropertyValues) {
+  updated(props: PropertyValues): void {
     if (props.has('searchTerm') && this.searchInput) {
       this.searchInput.value = this.searchTerm;
     }
@@ -100,6 +100,10 @@ export default class ExpandableSearchBar extends LitElement {
    */
   private clearSearch(): void {
     this.searchTerm = '';
+    /* istanbul ignore else */
+    if (this.searchInput) {
+      this.searchInput.focus();
+    }
     this.emitSearchClearedEvent();
   }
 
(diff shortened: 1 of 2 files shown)


packages/expandable-search-bar/src/expandable-search-bar.ts
```diff
@@ -29,6 +29,8 @@ export default class ExpandableSearchBar extends LitElement {
 
   @property({ type: Boolean }) showsDisclosure = false;
 
+  // TODO: searchTerm should be single source of state;
+  //       searchInput.value should not also affect state
```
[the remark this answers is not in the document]
hotzjacobb: Thanks; used in 247c417

packages/expandable-search-bar/src/expandable-search-bar.ts
```diff
@@ -80,6 +80,19 @@ export default class ExpandableSearchBar extends LitElement {
...
+   * Workaround for lit-element 2.2.1 not reflecting this.searchTerm 
+   * in the HTML element. updated is called when this.searchTerm's value
+   * is changed.
+   *
+   * @memberof ExpandableSearchBar
+   */
+  updated() {
+      if (this.searchInput) {
+        this.searchInput.value = this.searchTerm;
+      }
```
jbuckner, line 93: The `updated()` method gets called when any of the properties get updated so we'd want to limit this to just when `searchTerm` is updated. You can add the [`(props: PropertyValues)`](https://lit.dev/docs/components/lifecycle/#updated) argument to the method which contains all of the properties that were updated, ie `if (props.has('searchTerm')) { ... }`.

Also, could you reindent to 2 spaces?

packages/expandable-search-bar/src/expandable-search-bar.ts
```diff
@@ -80,6 +80,19 @@ export default class ExpandableSearchBar extends LitElement {
     `;
   }
 
+  /**
+   * Workaround for lit-element 2.2.1 not reflecting this.searchTerm
```
jbuckner, line 84: This isn't really a workaround so much as how form inputs work. The `searchTerm` property isn't bound to the input's value as a reactive property so you have to do set it when the value gets updated.

jbuckner (changes requested):

packages/expandable-search-bar/src/expandable-search-bar.ts
```diff
@@ -88,11 +100,6 @@ export default class ExpandableSearchBar extends LitElement {
    */
   private clearSearch(): void {
     this.searchTerm = '';
-    /* istanbul ignore else */
-    if (this.searchInput) {
-      this.searchInput.value = '';
-      this.searchInput.focus();
```
jbuckner, line 94: We want to still keep the `searchInput` in focus after clearing the search.

packages/expandable-search-bar/src/expandable-search-bar.ts
```diff
@@ -80,6 +81,17 @@ export default class ExpandableSearchBar extends LitElement {
     `;
   }
 
+  /**
+   * Update the DOM element after the value of `this.searchTerm` changes
+   *
+   * @memberof ExpandableSearchBar
+   */
+  updated(props: PropertyValues) {
```
jbuckner, line 89: Could you add a `: void` return type to this?

jbuckner (changes requested):
I tested the branch and the functionality looks good except for focusing the search box after clearing. Update that and add the return type to `updated()` and this PR should be good to go.

jbuckner (approved):
Looks good! Thank you for the fix!

PR #580, opened 2021-06-22
Remove ia-icons and mobile-top-nav

jbuckner:
The iaux icons now live in https://github.com/internetarchive/iaux-icons and `mobile-top-nav` is now part of `ia-topnav`


iisa (approved):
👍

PR #581, opened 2021-06-28
WEBDEV-4557 - Fix guest user dropdown position when there is no search bar in topnav

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-4557

### DIFF
--- a/packages/ia-topnav/src/ia-topnav.js
+++ b/packages/ia-topnav/src/ia-topnav.js
@@ -172,6 +172,7 @@ export default class IATopNav extends LitElement {
         .baseHost=${this.baseHost}
         .config=${this.config}
         .open=${this.signedOutOpened}
+        ?hideSearch=${this.hideSearch}
         tabindex="${this.signedOutTabIndex}"
         .menuItems=${this.signedOutMenuItems}
       ></signed-out-dropdown>
--- a/packages/ia-topnav/src/styles/signed-out-dropdown.js
+++ b/packages/ia-topnav/src/styles/signed-out-dropdown.js
@@ -5,13 +5,13 @@ export default css`
     .initial,
     .closed,
     .open {
-      right: 33.7rem;
+      right: 34rem;
     }
 
     .search-hidden.initial,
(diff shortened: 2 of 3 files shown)


packages/ia-topnav/src/ia-topnav.js
```diff
@@ -172,6 +172,7 @@ export default class IATopNav extends LitElement {
         .baseHost=${this.baseHost}
         .config=${this.config}
         .open=${this.signedOutOpened}
+        ?hideSearch=${this.hideSearch}
```
jbuckner, line 175: Q: Is this part of the fix or is this something else?

packages/ia-topnav/src/styles/signed-out-dropdown.js
```diff
@@ -5,27 +5,27 @@ export default css`
...
     .open {
-      right: 39.7rem;
+      right: 40rem;
     }
 
     .search-hidden.initial,
     .search-hidden.closed,
     .search-hidden.open {
-      right: 23.5rem;
+      right: 24.5rem;
```
jbuckner, line 28: Q: It seems like these adjustments are only a few pixels each, but the error looks like it's offset by 100-200px. How are they shifting that menu over so far?

jbuckner (approved):
Two questions, but I tested the demo and it looks good. 👍

jbuckner:
Oh I see what happened! The search wasn't showing when it was supposed to and we were offsetting assuming it was visible.

nsharma123:
> Oh I see what happened! The search wasn't showing when it was supposed to and we were offsetting assuming it was visible. Looks good

Yeah, the original fix is just passing `?hideSearch=${this.hideSearch}` prop in `<signed-out-dropdown>` component. The small change is  rem units just to show dropdown arrow more correctly.

packages/ia-topnav/src/ia-topnav.js
nsharma123: Yes

PR #583, opened 2021-07-02
Topnav optional second logo

iisa:
**Description**
Adds second logo next to parthenon as slot

**Technical**

it's a slotted piece that uses a config flag to open named slot.

**Testing**

pull down branch, npm install && npm start -> demo should show slotted link

**Evidence**
image

+ demo toggle
ezgif com-gif-maker(1)


packages/ia-topnav/src/primary-nav.js
```diff
@@ -116,6 +117,10 @@ class PrimaryNav extends TrackedElement {
     return this.openMenu === 'search';
   }
 
+  get allowSecondaryIcon() {
+    return this.config.secondIdentitySlot === 'allow';
```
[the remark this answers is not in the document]
iisa: moved it up 👍

packages/ia-topnav/src/ia-topnav.js
```diff
@@ -190,7 +191,12 @@ export default class IATopNav extends LitElement {
     return this.menus.more;
   }
 
+  get allowSecondaryIcon() {
+    return this.config.secondIdentitySlot === 'allow';
+  }
+
   render() {
+    const secondLogo = this.allowSecondaryIcon ? html`<slot name="opt-sec-logo"><slot>` : nothing;
```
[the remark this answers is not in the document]
iisa: hmm, i thought we found out that getters get computed at each update?
i do like the partition of this, moved to getters ^_^

jbuckner:
> i thought we found out that getters get computed at each update?

The computation is going to have to happen using either approach, but it's more of an issue when it's an expensive calculation. 

I did a small test to verify what DOM elements were being updated using getters vs binding and it was the same. When the number gets updated, only that DOM node got updated, but it left the `<p>BOOP</p>` untouched, even though that calculation was technically still happening.

https://user-images.githubusercontent.com/51138/124631873-a86a3680-de38-11eb-865e-01c757d403c2.mov

packages/ia-topnav/src/primary-nav.js
```diff
@@ -139,18 +145,30 @@ class PrimaryNav extends TrackedElement {
...
   }
 
+  get secondLogoSlot() {
+    return this.allowSecondaryIcon
+      ? html`<slot name="opt-sec-logo" data-event-click-tracking="${this.config.eventCategory}|${this.secondIdentitySlot}"><slot>`
+      : nothing;
+  }
+
+  get logoSubtext() {
+    return this.allowSecondaryIcon ? '' : logoWordmark;
```
jbuckner, line 155: Q: Can the `''` be `nothing`?

packages/ia-topnav/src/primary-nav.js
```diff
@@ -139,18 +145,30 @@ class PrimaryNav extends TrackedElement {
     `;
   }
 
+  get secondLogoSlot() {
+    return this.allowSecondaryIcon
+      ? html`<slot name="opt-sec-logo" data-event-click-tracking="${this.config.eventCategory}|${this.secondIdentitySlot}"><slot>`
```
jbuckner, line 150: Q: Does this click tracking work? I was expecting the link inside it to capture the click.

jbuckner (approved):
A few comments and suggestions, but looks good overall

packages/ia-topnav/src/ia-topnav.js
```diff
@@ -51,6 +52,7 @@ export default class IATopNav extends LitElement {
       username: { type: String },
       userProfileImagePath: { type: String },
       userProfileLastModified: { type: String },
+      secondIdentitySlot: { type: String },
```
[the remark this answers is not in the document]
iisa: good call. I will rename.

packages/ia-topnav/src/primary-nav.js
```diff
@@ -116,6 +118,10 @@ class PrimaryNav extends TrackedElement {
     return this.openMenu === 'search';
   }
 
+  get allowSecondaryIcon() {
+    return this.secondIdentitySlot === 'allow';
```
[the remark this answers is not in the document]
iisa: mmm I believe I have this as an attribute served by PHP.. let's keep to "secondIdentitySlotMode"

packages/ia-topnav/src/primary-nav.js
iisa: thanks for catching!

packages/ia-topnav/src/primary-nav.js
iisa: Yano, idk! this was a hopeful add-on ^_^

this `data-` attribute connects to the 1% GA instance, I can add a click handler to see if it reacts to a click event...

iisa: yah, no dice. yanking.

PR #586, opened 2021-07-30
Offshoot Topnav Updates

jbuckner:
Update Typescript declarations with some new attributes and fix some existing menu options. There were also some z-index issues when the menu was supposed to be floating over other elements on screen.


iisa (approved):
👍

PR #606, opened 2021-11-01
keep minutes display string through crunching

iisa:
- update minutesDisplay to keep type through processing, from [code review](https://github.com/internetarchive/iaux/pull/605#discussion_r740340847)

### DIFF
--- a/packages/ia-components/components/theatres/with-youtube-spotify/utils/get-track-list-by-source.js
+++ b/packages/ia-components/components/theatres/with-youtube-spotify/utils/get-track-list-by-source.js
@@ -9,12 +9,9 @@ const formatTime = (lengthInSeconds) => {
   const seconds = ((lengthInMS % 60000) / 1000).toFixed(0); // becomes a string, thanks JS math
   const extraMinute = seconds === '60';
   const secondsDisplay = extraMinute ? '00' : (seconds < 10 ? '0' : '') + seconds;
-  let minutesDisplay = minutes;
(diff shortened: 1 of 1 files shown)


packages/ia-components/components/theatres/with-youtube-spotify/utils/get-track-list-by-source.js
```diff
@@ -9,12 +9,9 @@ const formatTime = (lengthInSeconds) => {
   const seconds = ((lengthInMS % 60000) / 1000).toFixed(0); // becomes a string, thanks JS math
   const extraMinute = seconds === '60';
   const secondsDisplay = extraMinute ? '00' : (seconds < 10 ? '0' : '') + seconds;
-  let minutesDisplay = minutes;
-  if (!minutesDisplay) {
-    minutesDisplay = '00';
-  }
+  let minutesDisplay = minutes ? `${minutes}` : '00';
```
jbuckner, line 12: One minor update to this that I didn't realize in the other comment is that you're using the falsey check to check if `minutes` is `0`, correct (ie not `undefined`)? 

I might update this check to be `minutes === 0` if so. The `minutes + 1` below got me wondering if `minutes` could be `undefined` because of the falsey check.
```suggestion
  let minutesDisplay = minutes === 0 ? `${minutes}` : '00';
```

jbuckner (approved):
One suggestion, but overall looks good and thanks for keeping up the great coding practices!

PR #616, opened 2022-01-05
liner notes player: use image in item in media window instead of image placeholder

iisa:
**Description**
Adds safety net to use floating/single image as cover for media window

Demo updates
- use list to draw table
- smol table redesign
- new stub for 78 item w/ 1 image

### DIFF
--- a/packages/ia-components/components/theatres/with-youtube-spotify/utils/archive-default-album-parser.js
+++ b/packages/ia-components/components/theatres/with-youtube-spotify/utils/archive-default-album-parser.js
@@ -39,7 +39,7 @@ const archiveDefaultAlbumParser = ({ fileDirectoryPrefix, files }) => {
     const isOriginal = source === 'original';
     const isAudioFile = isValidAudioFile(currentFileName);
     const isItemImageFile = isOriginal && isValidImageFile(currentFileName);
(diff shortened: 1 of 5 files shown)


packages/ia-components/components/theatres/with-youtube-spotify/utils/archive-default-album-parser.js
```diff
@@ -39,10 +39,11 @@ const archiveDefaultAlbumParser = ({ fileDirectoryPrefix, files }) => {
     const isOriginal = source === 'original';
     const isAudioFile = isValidAudioFile(currentFileName);
     const isItemImageFile = isOriginal && isValidImageFile(currentFileName);
-
+    const isThumbnail = !!currentFileName.match(/__ia_thumb.jpg/g);
```
mmcnl, line 42: Hi Isa! Would it make sense to anchor this regex to the end of the string to make it a bit more strict, eg `/__ia_thumb.jpg$/g`?

mmcnl:
@iisa  LGTM!

packages/ia-components/components/theatres/with-youtube-spotify/utils/archive-default-album-parser.js
iisa: hi matt, thanks for the suggestion. updated.

PR #748, opened 2022-08-10
WEBDEV-5404 Fix attributes option in lazy load call

latonv:
A call to lazy-load-service's loadScript method was using `[{key: 'k', value: 'v'}]` format for its `attributes` option.
The latest lazy-load-service interface expects a single Record object containing all the attributes, i.e. `{k: 'v'}`.

**Description**

> What does this PR achieve? Is it a feature/hotfix/refactor? Does it close an Issue? If so, use `Closes #`.

The Zendesk widget on Offshoot pages is currently breaking due to this loadScript call. This PR should resolve the issue and allow the help widgets to become functional again.

**Technical**

Small one-liner fix, just adjusting how the help widget passes attributes to the lazy-loader.

**Testing**

TBD

**Evidence**


packages/ia-zendesk-help-widget/package.json
```diff
@@ -28,7 +28,7 @@
     "storybook:build": "build-storybook -o _site -s storybook-static"
   },
   "dependencies": {
-    "@internetarchive/lazy-loader-service": "^0.1.0",
+    "@internetarchive/lazy-loader-service": "^0.2.0",
```
latonv, line 31: This is required for the Offshoot pages, but I'm unsure whether this upgrade will cause issues with the widget on pages that aren't in Offshoot yet. Do those pages still require the older version?

iisa (approved):
LGTM!

packages/ia-zendesk-help-widget/package.json
latonv: Update: shouldn't be an issue -- petabox will continue using the older widget version.

iisa (approved):
:lgtm:

PR #756, opened 2022-10-26
WEBDEV-5525: Direct topnav search to legacy/beta search according to opt-in

latonv:
**Description**

Updates the ia-topnav component to respect the user's opt-in/opt-out preference for the search beta, ensuring that searches from the topnav's search bar go to the correct search page. Also slightly updates the search option labels to match how they are written on the actual search page widget.

**Technical**

Uses the `SearchBeta-opt-in` localStorage key that gets set upon clicking legacy/beta search links.

**Testing**

- For a user in the search beta segment who has opted in by clicking a "Try the new search beta" link, performing a search from the topnav should direct to the beta search page (`/search`).
- For users who either aren't in the search beta, or haven't yet opted into beta search, performing a search from the topnav should direct to the legacy search page (`/search.php`).

### DIFF
--- a/packages/ia-topnav/src/nav-search.js
+++ b/packages/ia-topnav/src/nav-search.js
@@ -29,6 +29,9 @@ class NavSearch extends TrackedElement {
     this.open = false;
     this.openMenu = '';
     this.searchIn = '';
+    this.inSearchBeta = false;
+
+    this.initSearchBetaOptIn();
   }
 
   updated() {
@@ -36,6 +39,14 @@ class NavSearch extends TrackedElement {
       this.shadowRoot.querySelector('[name=query]').focus();
     }
     return true;
+  }
+
+  initSearchBetaOptIn() {
+    try {
+      this.inSearchBeta = JSON.parse(window.localStorage?.getItem('SearchBeta-opt-in'));
+    } catch (err) {
+      // This is fine, as the key may simply not exist
+    }
   }
 
   search(e) {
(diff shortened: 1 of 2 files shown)


packages/ia-topnav/src/nav-search.js
```diff
@@ -38,6 +41,14 @@ class NavSearch extends TrackedElement {
     return true;
   }
 
+  initSearchBetaOptIn() {
+    try {
+      this.inSearchBeta = JSON.parse(window.localStorage?.getItem('SearchBeta-opt-in'));
+    } catch (err) {
```
traceypooh, line 47: whoa localStorage instead of cookies, eh?  neat. :)

packages/ia-topnav/src/nav-search.js
```diff
@@ -74,6 +85,10 @@ class NavSearch extends TrackedElement {
     return this.searchIn ? html`<input type='hidden' name='sin' value='${this.searchIn}' />` : nothing;
   }
 
+  get searchEndpoint() {
+    return this.inSearchBeta ? '/search' : '/search.php';
+  }
```
traceypooh, line 90: lol

packages/ia-topnav/src/nav-search.js
latonv: Yeah, I figure since the opt in/out preference is entirely a client-side affair there's little reason for it to be sent as a cookie with every request! (The AB test system in petabox has no knowledge of opt in/out, only beta link visibility)

bfalling (approved):
LGTM

iisa (approved):

PR #760, opened 2022-11-01
fix for rendertron-ed nav input type-in colors

traceypooh:
**Description**

> fixes archive.org non-offshoot-served pages nav form typeins from white-on-white

**Technical**

> seems like rendertron-created CSS causes browser defaults of `color:fieldtext` implicitly relied on rule to not work.  so need to get explicit

**Testing**

> IDK.  dont have spoons to push 3 MRs, npm publish, offshoot package.json version, mr, push, and pi web -- so would like to get this thru from bottom up and i'll get more formal if this fix doesnt work (though it should!) 

**Evidence**

### DIFF
--- a/packages/ia-topnav/src/styles/ia-topnav.js
+++ b/packages/ia-topnav/src/styles/ia-topnav.js
@@ -58,10 +58,6 @@ export default css`
     font-family: var(--themeFontFamily);
   }
 
-  input {
-    color: var(--grey13);
-  }
-
   primary-nav:focus {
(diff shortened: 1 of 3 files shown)


traceypooh:
NOTE: i *did* simulate this as a fix on petabox top-down `archive.less` similar fix.  but that's heavyhanded and really the nav repo should "own" its colors.

traceypooh:
image

latonv (approved):

PR #809, opened 2022-12-21
Fix search-menu overlapped with donation banner in offshoot

dualcnhq:
**Description**

> `Closes WEBDEV-5703`.

**Technical**

> updated CSS var `--topOffset` value

**Testing**

> see WEBDEV-5703

**Evidence**

> see WEBDEV-5703

### DIFF
--- a/packages/ia-topnav/src/styles/search-menu.js
+++ b/packages/ia-topnav/src/styles/search-menu.js
@@ -2,7 +2,7 @@ import { css } from 'https://offshoot.ux.archive.org/lit.js';
 
 export default css`
   :host {
-    --topOffset: 1200px;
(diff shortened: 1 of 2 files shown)


packages/ia-topnav/src/styles/search-menu.js
```diff
@@ -2,7 +2,7 @@ import { css } from 'https://offshoot.ux.archive.org/lit.js';
 
 export default css`
   :host {
-    --topOffset: -800px;
+    --topOffset: 1200px;
```
latonv, line 5: Q: Do we want -1200 here? (Pushing it up rather than down)

latonv (approved):
LGTM, just one possible typo.

PR #830, opened 2023-01-11
Lazy load topnav images

cdrini:
**Description**

Remove 10 network requests of images we might not need:

image

**Technical**

**Testing**

I haven't tested this... should hopefully do the trick, but at worst it'll just be the same as it was before.

**Evidence**


iisa (approved):
looks good.

PR #859, opened 2023-05-22
Define activity indicator in main indexjs

iisa:
## Problem:
our esm instance cannot reconcile a sibling file of its main index.js with the same name as the package.

problem file for esm: https://github.com/internetarchive/iaux/blob/master/packages/ia-activity-indicator/ia-activity-indicator.js
it thinks it's https://github.com/internetarchive/iaux/blob/master/packages/ia-activity-indicator/index.js

See: https://esm.archive.org/v99/@internetarchive/ia-activity-indicator/ia-activity-indicator.js
versus: https://unpkg.com/@internetarchive/ia-activity-indicator@0.0.4/ia-activity-indicator.js

## Solution:
define the custom element in the index.js file

dev note: we cannot define in the src/ia-activity-indicator.js file for backwards compatibility.
- if consumer is referring to ./ia-activity-indicator.js, it will create a double registry conflict.

## Testing:
pull down branch, run:
- npm i
- npm start

demo loads with activity indicator showing
image


traceypooh:
Q?  wouldn't we normally like the customElements define in the main .js file?

for example, we see this here:
```
tracey@minti  4:47PM  /Volumes/bff/dev/iaux/packages/ia-activity-indicator
> cat ia-activity-indicator.js 
import { IAActivityIndicator } from './src/ia-activity-indicator.js';

window.customElements.define('ia-activity-indicator', IAActivityIndicator);
```

but all the code is in 
`./src/ia-activity-indicator.js';`


so i'm curious why this got split this way?

traceypooh:
code & solve seem acceptable, but the more i look/poke, the more it seems like the 
```
customElements.define('ia-activity-indicator', IAActivityIndicator);
```

should just be added to the bottom of this file.
```
iaux/packages/ia-activity-indicator/src/ia-activity-indicator.js
```

iisa:
> code & solve seem acceptable, but the more i look/poke, the more it seems like the
> 
> ```
> customElements.define('ia-activity-indicator', IAActivityIndicator);
> ```
> 
> should just be added to the bottom of this file.
> 
> ```
> iaux/packages/ia-activity-indicator/src/ia-activity-indicator.js
> ```

yes, i agree. once i add it the definition to `src/ia-activity-indicator.js` -> all the consumers of `./ia-activity-indicator.js` will hit the double registry conflict.  then i will have to update each consumer just for this patch. this is the legacy snowpackJS pattern we used to use, which is now obsolete.  i am just fixing esm's issue, where it cannot find `./ia-activity-indicator.js`. 👍

traceypooh:
ok, it's merged.  but am curious for in future, what if we simply wrapped w/ a try/catch to avoid any double includes?
eg:
```js
try {
   customElements.define('ia-activity-indicator', IAActivityIndicator);
} catch {}
```

PR #867, opened 2023-08-29
Suppress admin links in topnav on homepage and search page

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-6330

### DIFF
--- a/packages/ia-topnav/src/ia-topnav.js
+++ b/packages/ia-topnav/src/ia-topnav.js
@@ -43,7 +43,6 @@ export default class IATopNav extends LitElement {
       /** Identifier for the item or collection currently being viewed */
       itemIdentifier: { type: String },
       mediaSliderOpen: { type: Boolean },
-      suppressAdminLinks: { type: Boolean },
       menus: {
         type: Object,
         converter(value) {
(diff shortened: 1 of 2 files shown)


packages/ia-topnav/src/ia-topnav.js
```diff
@@ -220,13 +222,13 @@ export default class IATopNav extends LitElement {
...
     const basicItems = this.menus.user;
-    
+
     let adminItems = this.menus.userAdmin;
     if (this.canManageFlags) {
       adminItems = adminItems.concat(this.menus.userAdminFlags);
     }
 
-    return this.admin
+    return this.admin && !this.suppressAdminLinks
```
latonv, line 231: Q: Instead of adding new state for suppression, would it work to just check whether `this.itemIdentifier` is set? I believe that property should be undefined on non-item/non-collection pages already.

latonv (commented):
Thanks for the fix for this! I _think_ there may be a simpler alternative to try here and avoid adding new state (see comment).

packages/ia-topnav/src/ia-topnav.js
nsharma123: Addressed your suggestion

nsharma123: Applied the above suggestion and working great!

latonv (approved):
LGTM, thanks for looking into the suggestion!

PR #873, opened 2023-10-10
WEBDEV-6423 User lists service migration

latonv:
**Description**

> What does this PR achieve? Is it a feature/hotfix/refactor? Does it close an Issue? If so, use `Closes #`.

Migrates the Offshoot user lists service into the `iaux-userlists-settings` package.

**Technical**

The existing component is modified to use the service calls rather than hitting the backend endpoint directly. Moreover, the output of the `userListSaved` event is changed to provide only the final JSON result of the service call, not a pair of HTTP Request/Response objects.

**Testing**

See ticket for QA steps.

**Evidence**


packages/ia-userlist-settings/src/user-lists-service/user-lists-service-interface.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Fetches a specific user list, including the list metadata and the array of list
+   * members. Note that the list members returned include only the basic information
+   * kept in the user lists store, not the full item metadata. For that, use `fetchListMembers`.
+   *
+   * @param userId The ID of the user whose list should be fetched
+   * @param listId The ID of the list to fetch (with respect to the given user)
+   */
+  fetchList(
```
bfalling, line 33: MINOR SUGGESTION: Consider `fetchListForUser` as the name to parallel `fetchAllListsForUser` and cuz in both cases, the first parameter is the `userId` so the function name reads well going into its first parameter.

bfalling: OTOH: Maybe you want the API to be as "list" focused as possible. (Cuz maybe the same argument applies to `fetchListMembers`.

bfalling (commented):
Visual inspection didn't reveal any blockers. Not sure if there are (or will be) tests for functionality.

packages/ia-userlist-settings/src/user-lists-service/user-lists-service-interface.ts
nsharma123: I would propose the following names for the functions.
`FetchLists()` -> Fetch all lists
`FetchList()` -> Fetch single list
`FetchListMembers()` 
`FetchListsForItem()` -> fetch lists for details page to shows if item is already in xxx list

packages/ia-userlist-settings/src/user-lists-service/user-lists-service.ts
```diff
@@ -0,0 +1,236 @@
...
+    this.baseUrl = baseUrl;
+  }
+
+  private async fetchEndpoint<T>(
+    url: string,
+    method?: string,
+    body?: BodyInit,
+  ): Promise<Result<T, UserListsError>> {
+    try {
+      const fetchResult: BackendServiceResponse<T> = await this.fetchHandler.fetchApiResponse(
```
nsharma123, line 69: How the `this.fetchHandler` being used in this repo? are you creating new handler we already in offshoot repo?

packages/ia-userlist-settings/src/user-lists-service/user-lists-service-interface.ts
latonv: I suppose the question is whether the `ForUser` bit adds anything useful.

My original thinking around this was to avoid having two different method names differing by a single `s`, and just be explicitly clear about which one is user-encompassing instead. The code will be read far more often than written after all, and relying on a one-letter difference to signify a large difference in functionality sets off alarm bells for me.

But on the other hand, I do agree that it feels slightly redundant given that these are already explicitly _user_ lists.

packages/ia-userlist-settings/src/user-lists-service/user-lists-service.ts
latonv: Good question -- plan is to extract Offshoot's fetch handler into iaux as well, so that it can be used here. However, the interface here only requires the one `fetchApiResponse` method and is not tied to a specific implementation, so in theory one could just pass in a simple handler that just delegates straight to the browser's `fetch`.

packages/ia-userlist-settings/src/user-lists-service/user-lists-service-interface.ts
latonv: Had some in-person chat about this, and I think we settled on the following:

- `fetchListsForUser(userId)` -- returns all lists owned by the given user
- `fetchList(userId, listId)` -- returns a single, specific list, as identified by the `userId`/`listId` pair
- `fetchListMembers(userId, listId)` -- returns the array of members for a specific list
- `fetchOwnListsContainingItem(itemId)` -- returns all of the _current_ user's lists which contain the given item identifier

The rationale is a mixture of clarity and future-proofing. We are likely to eventually extend this service with methods to fetch lists across _multiple_ users. If we reserve a very general method name like `fetchLists` for the _user-specific_ method now, it could hamper us in the future to be stuck with that among _even more general_ cases that have to tiptoe around the existing name. Better to be explicit & clear about what the method does.

Likewise, `fetchOwnListsContainingItem` makes it clear upfront that the method acts on the _current user's_ lists, and allows us to cleanly extend the service with methods that act on a more general set of users. The name `fetchListsForItem` doesn't fully convey these semantics.

PR #875, opened 2023-10-11
user list settings: disable save button on modal on click, re-enable if error happens

iisa:


packages/ia-userlist-settings/src/ia-user-list-settings.ts
```diff
@@ -40,6 +40,8 @@ export class IAUserListSettings extends LitElement {
 
   private async saveListDetails(event: Event) {
     event.preventDefault();
+    const submitButton = (event.target as HTMLElement)?.querySelector('button#save-list-settings');
+    submitButton?.setAttribute('disabled', 'true');
```
bfalling, line 44: QUESTION: If one re-opens the "Add to List" modal later on the same Details page (for example, after the user has submitted a list request which completes successfully), will the `disabled` attribute be cleared?

bfalling (commented):
Just one question; otherwise, LGTM.

packages/ia-userlist-settings/src/ia-user-list-settings.ts
iisa: hi yes, that's working as expected.

iisa:
closing to use https://github.com/internetarchive/iaux/pull/877

packages/ia-userlist-settings/src/ia-user-list-settings.ts
```diff
@@ -40,6 +40,8 @@ export class IAUserListSettings extends LitElement {
 
   private async saveListDetails(event: Event) {
     event.preventDefault();
+    const submitButton = (event.target as HTMLElement)?.querySelector('button#save-list-settings');
```
nsharma123, line 43: Instead of playing with with DOM, can't we simply use lit state to disable the button after initial click?

iisa: ideally, that's the route. hope we can fix it next pass

PR #877, opened 2023-10-13
Userlists setting modal - Prevent multiple clicks on Save button

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-6455


packages/ia-userlist-settings/src/ia-user-list-settings.ts
```diff
@@ -38,10 +44,13 @@ export class IAUserListSettings extends LitElement {
 
   @query('#private') private listPrivate: HTMLInputElement;
 
+  @state() private disableSaveButton = false;
```
nsharma123, line 47: This is specific change to disable save button after initial click and enable if any error occure.

PR #905, opened 2024-04-16
IA-Topnav unit conversion from px to rem to make responsive in-respec…

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-6784

### DIFF
--- a/packages/ia-topnav/package.json
+++ b/packages/ia-topnav/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/ia-topnav",
-  "version": "1.3.0",
+  "version": "1.3.1-alpha7",
   "description": "Top nav for Internet Archive",
   "license": "AGPL-3.0-only",
   "main": "index.js",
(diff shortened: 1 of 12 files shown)


packages/ia-topnav/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/ia-topnav",
-  "version": "1.3.0",
+  "version": "1.3.1-alpha7",
```
nsharma123, line 3: this will be removed after final CR.

packages/ia-topnav/src/styles/media-button.js
```diff
@@ -43,6 +41,10 @@ export default css`
     -ms-flex-pack: center;
     justify-content: center;
   }
+  .menu-item > .icon > svg {
+    height: 4rem;  
+    width: 4rem;
```
iisa, line 46: 4rem is prominent i've seen this in multiple areas, can we make this a css var with 4rem default?

iisa (approved):
ok, given it passes qa. could use some high level css var grouping around shared sizes if possible.

PR #916, opened 2024-06-18
Implement design for keyboard control of upper navigational area

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-6829

### DIFF
--- a/packages/ia-topnav/src/lib/keyboard-navigation.js
+++ b/packages/ia-topnav/src/lib/keyboard-navigation.js
@@ -1,0 +1,130 @@
+
+
+
+export default class KeyboardNavigation {
+  /**
+   * Constructor for the KeyboardNavigation class.
+   * @param {HTMLElement} elementsContainer - The container element that holds the focusable elements.
+   * @param {string} menuOption - The type of menu option ('web' or 'usermenu').
+   */
+  constructor(elementsContainer, menuOption) {
+    this.elementsContainer = elementsContainer;
+    this.menuOption = menuOption;
+    this.focusableElements = this.getFocusableElements();
+    this.focusedIndex = this.getInitialFocusedIndex();
+
+    this.focusableElements[this.focusedIndex]?.focus();
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+  }
+
+  /**
+   * Returns the initial focused index based on the menu option.
+   * @returns {number} The initial focused index (0 for 'web', 1 for 'usermenu').
+   */
+  getInitialFocusedIndex() {
+    return this.menuOption === 'usermenu' ? 1 : 0;
+  }
+
+  /**
+   * Gets an array of focusable elements within the container.
+   * @returns {HTMLElement[]} An array of focusable elements.
+   */
+  getFocusableElements() {
+    const focusableTagSelectors = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
+    const isDisabledOrHidden = el => !el.hasAttribute('disabled') && !el.getAttribute('aria-hidden');
+
+    let elements;
+    if (this.menuOption === 'web') {
+      const waybackSlider = this.elementsContainer.querySelector('wayback-slider').shadowRoot;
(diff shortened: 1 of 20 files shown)


packages/ia-topnav/src/lib/keyboard-navigation.js
```diff
@@ -0,0 +1,128 @@
...
+   */
+  focusNextMenuItem(isPrevious = false) {
+    this.elementsContainer.dispatchEvent(
+      new CustomEvent('focusToNext', {
+        bubbles: true,
+        composed: true,
+        detail: {
+          mediatype: this.menuOption,
+          moveTo: isPrevious ? 'prev' : 'next',
+        },
```
latonv, line 124: I find the naming a bit inconsistent here, since above we have separate `focusNext` and `focusPrevious` methods, yet here we have only a single `focusNextMenuItem` method which handles both directions. I'd argue that consistency & clarity should take precedence over conciseness here, though perhaps only a minor concern.

packages/ia-topnav/src/media-button.js
```diff
@@ -103,6 +103,7 @@ class MediaButton extends TrackedElement {
         @click=${this.followable ? this.trackClick : this.onClick}
         data-event-click-tracking="${this.analyticsEvent}"
         title="${this.tooltipPrefix} ${this.mediatype} menu"
+        tabindex="${this.openMenu === 'media' ? '' : '0'}"
```
latonv, line 106: Q: Does this approach add and remove `tabindex` from elements dynamically?

packages/ia-topnav/src/media-menu.js
```diff
@@ -75,6 +76,25 @@ class MediaMenu extends LitElement {
...
+      const mediaButtons = Array.from(this.shadowRoot.querySelectorAll('media-button'));
+
+      mediaButtons.map((button, index) => {
+        const linkItem = button.shadowRoot.querySelector('a.menu-item');
+        if (linkItem) {
+          if (linkItem.classList.contains(`${this.selectedMenuOption}`)) {
+            linkItem.classList.remove('selected');
+            linkItem.blur();
+
+            mediaButtons[this.currentTab.moveTo === 'next' ? index + 1 : index - 1].shadowRoot.querySelector('a.menu-item').focus();
```
latonv, line 93: Just for ease of reading & avoiding overly long lines, could we do something like:

```suggestion
            const newFocusIndex = this.currentTab.moveTo === 'next' ? index + 1 : index - 1;
            mediaButtons[newFocusIndex].shadowRoot.querySelector('a.menu-item').focus();
```

latonv (approved):
LGTM, just had a few small questions/suggestions, no blockers found. Nice work on this!

packages/ia-topnav/src/lib/keyboard-navigation.js
```diff
@@ -0,0 +1,128 @@
...
+   * @param {string} menuOption - The type of menu option ('web' or 'usermenu').
+   */
+  constructor(elementsContainer, menuOption) {
+    this.elementsContainer = elementsContainer;
+    this.menuOption = menuOption;
+    this.focusableElements = this.getFocusableElements();
+    this.focusedIndex = this.getInitialFocusedIndex();
+
+    this.focusableElements[this.focusedIndex]?.focus();
+    this.handleKeyDown = this.handleKeyDown.bind(this);
```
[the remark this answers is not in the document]
nsharma123: Basically this method bind the any key click event from keyboard. I am binding this event for subnav pane and usermenu to move focus on elements by arrow keys.

packages/ia-topnav/src/lib/keyboard-navigation.js
nsharma123: Make sense, I am changing function name. BTW, this function is to move focus on other parent items, where the above function moves focus in within a container like subnav or usermenu element

packages/ia-topnav/src/media-button.js
nsharma123: Its seems not usable change. reverting this. Thanks!

PR #917, opened 2024-06-27
[ia-topnav] WEBDEV-6909 Remove old references to search.php (now /search) and obsolete beta flag

latonv:
Converts old references to the `/search.php` endpoint to `/search`, so that searches are sent directly to the newer page without an extra redirect. Also eliminates a related obsolete flag for the search beta opt-in, which is no longer required.


iisa (approved):
👍

PR #924, opened 2024-11-21
[ia-userlist-settings] WEBDEV-7130 Send appropriate content-type header on lists CRUD requests

latonv:
Ensures that User Lists methods that send JSON bodies set their `Content-Type` header correctly.


bfalling (approved):
LGTM

PR #932, opened 2025-01-14
WEBDEV-7257: Fix ia-activity-indicator double-registration

jbuckner:
Currently we are registering the ia-activity-indicator in two different locations so depending on how you import it, you can potentially get an error that it has already been registered.


packages/ia-activity-indicator/index.js
```diff
@@ -1,5 +1,5 @@
-import { IAActivityIndicator } from './src/ia-activity-indicator.js';
+import { IAActivityIndicator, IAActivityIndicatorMode } from './src/ia-activity-indicator.js';
 
 window.customElements.define('ia-activity-indicator', IAActivityIndicator);
 
-export { IAActivityIndicator, IAActivityIndicatorMode } from './src/ia-activity-indicator.js';
+export { IAActivityIndicator, IAActivityIndicatorMode };
```
latonv, line 5: Q: I'm mostly just curious why we needed to register the element here in `index.js` at all. Shouldn't it get registered automatically by virtue of extending `LitElement`?

latonv (approved):
LGTM, just a bit of curiosity about the approach this package is using.

packages/ia-activity-indicator/index.js
jbuckner: In our TS repos, you can register with the `@customElement('ia-activity-indicator')` decorator, but this is a JS package so you have to manually register with `window.customElements.define('ia-activity-indicator', IAActivityIndicator);`

This is a pretty old Lit package (maybe one of our first?) and they used to recommend registering the component independently of the actual component definition so we had this separate registration file. Now it's usually just bundled together in the same file.

jbuckner: I moved the registration into the element itself so we get the registration regardless of how it gets imported.

PR #935, opened 2025-01-20
Top Nav navigation disappears when resizing browser

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-7230

### DIFF
--- a/packages/ia-topnav/src/styles/primary-nav.js
+++ b/packages/ia-topnav/src/styles/primary-nav.js
@@ -56,8 +56,8 @@ export default css`
   }
 
   media-menu {
-    grid-column-start: hamburger-start;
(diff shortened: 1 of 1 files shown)


packages/ia-topnav/src/styles/primary-nav.js
```diff
@@ -56,8 +56,8 @@ export default css`
   }
 
   media-menu {
-    grid-column-start: hamburger-start;
-    grid-column-end: user-end;
+    flex: 1;
```
rebecca-shoptaw, line 59: Nice! This approach should be much more flexible (so to speak)

rebecca-shoptaw (approved):
LGTM!

PR #936, opened 2025-01-23
WEBDEV-7374: Update search-service in user list

jbuckner:


rebecca-shoptaw (commented):
Same question here as the other PR -- is it just a `package.json` update for the search service plus the linter config updates, or is there something else in there you'd like people to take a look at?

rebecca-shoptaw (commented):
I think we're trying to avoid pushing up these alpha-bump commits in general so that there's no need to go back and remove them before merge -- but as long as it gets removed this is fine!

jbuckner:
@rebecca-shoptaw Yeah the main update here is the search service bump in `package.json`. It really should have no impact on consumers. I'm just trying to get us off of the 1.x branch of search service since it had the extra copy of all of the metadata fields.

I went ahead and updated the repo with newer versions of the dev dependencies (ie eslint) so it should more closely match the TS WC template now.

rebecca-shoptaw (approved):

PR #939, opened 2025-02-24
Fixes for ia-pic-uploader

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-7282

### DIFF
--- a/packages/ia-pic-uploader/src/ia-pic-uploader.ts
+++ b/packages/ia-pic-uploader/src/ia-pic-uploader.ts
@@ -1,10 +1,18 @@
-import { html, css, LitElement, CSSResultGroup, nothing, svg } from 'lit';
+import {
+  html,
+  LitElement,
+  CSSResultGroup,
+  nothing,
+  svg,
+  TemplateResult,
+} from 'lit';
 import { property, customElement, state, query } from 'lit/decorators.js';
-import iaButtonStyle from './style/ia-button-style';
 import { BackendServiceHandler } from './services/backend-service';
-import log from './services/log';
-
+import { ServiceOptionType, ValidFileType, ValidType } from './model';
 import '@internetarchive/ia-activity-indicator/ia-activity-indicator';
+import iaButtonStyle from './style/ia-button-style';
+import iaPicUploaderStyle from './style/ipic-uploader-style';
+import log from './services/log';
 
 @customElement('ia-pic-uploader')
 export class IAPicUploader extends LitElement {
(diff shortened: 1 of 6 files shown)


packages/ia-pic-uploader/src/ia-pic-uploader.ts
```diff
@@ -368,70 +351,81 @@ export class IAPicUploader extends LitElement {
...
+        if (this.type === 'full') await this.metadataAPIExecution();
+        log('file saved, metadata call started');
+      } else {
+        log('Failed to save file', saveResponse);
+      }
+    } catch (err) {
+      log('Failed to save file', err);
+      this.showLoadingIndicator = false;
+    } finally {
+      if (this.type === 'compact') this.showLoadingIndicator = false;
```
jbuckner, line 380: Q: In this `finally` block, you're checking for the type before hiding the loading indicator, but in the `catch` block above, you're not. How come the different approaches?

packages/ia-pic-uploader/src/ia-pic-uploader.ts
```diff
@@ -368,70 +351,81 @@ export class IAPicUploader extends LitElement {
...
   /**
    * after upload, verify using metadata API if successfully uploaded or not
-   *
-   * @memberof IAPicUploader
    */
-  metadataAPIExecution() {
+  async metadataAPIExecution() {
     const now = Math.round(Date.now() / 1000); // like unix time()
 
     const metadataApiInterval = setInterval(async () => {
```
jbuckner, line 393: I'm not familiar with how the uploader works, but what is this `setInterval()` doing?

packages/ia-pic-uploader/src/ia-pic-uploader.ts
```diff
@@ -368,70 +351,81 @@ export class IAPicUploader extends LitElement {
...
+    }
   }
 
   /**
    * after upload, verify using metadata API if successfully uploaded or not
-   *
-   * @memberof IAPicUploader
    */
-  metadataAPIExecution() {
+  async metadataAPIExecution() {
```
jbuckner, line 390: What is happening async in this method?

packages/ia-pic-uploader/src/services/backend-service.ts
jbuckner, line 1: What was the motivation behind removing this service? It seems like the UI component is now doing quite a bit with the UI and backend/network communications.

packages/ia-pic-uploader/src/ia-pic-uploader.ts
```diff
@@ -202,7 +192,7 @@ export class IAPicUploader extends LitElement {
     document.addEventListener('dragover', e => this.dragOver(e), false);
     document.addEventListener('dragleave', e => this.dragLeave(e), true);
     document.addEventListener('drop', e => this.drop(e), false);
-    document?.addEventListener('saveProfileAvatar', (e: Event) => {
+    document.addEventListener('saveProfileAvatar', (e: Event) => {
```
jbuckner, line 195: What is emitting this event?

jbuckner: Is the `/account/settings` page emitting that event, then this component is listening for that event to start the saving process?

If that is the case, could we just call a method directly on this component since the settings page consumes this component? The event approach seems a bit roundabout way of communicating this intent.

PR #943, opened 2025-03-13
WEBDEV-7370 Added shared iaStyles

nsharma123:
**Description**
We've been using the `ia-button.ts` file to maintain a consistent design experience for buttons across various states (:hover, :active, :selected, etc.). However, this requires duplicating the file in multiple repositories (ia-book-actions, bookreader, offshoot, collection-browser, account-settings, etc.). Moreover, some repositories don't even utilize color effects on buttons.

Fo fix the issue, publishing a package called `@internetarchive/ia-styles`, which would contain these shared styles. This way, we can simply import and use them

https://webarchive.jira.com/browse/WEBDEV-7370


nsharma123:
This is how demo page looks like on local server: 
image

packages/ia-styles/demo/app-root.ts
```diff
@@ -0,0 +1,60 @@
...
+      <table class="initial">
+        <tr>
+          <td>
+            <button class="ia-button dark">Dark Button</button>
+            <button class="ia-button primary">Primary Button</button>
+            <button class="ia-button danger">Danger Button</button>
+            <button class="ia-button warning">Warning Button</button>
+            <button class="ia-button" disabled>Disabled</button>
+            <a href="#" class="ia-button primary"
+              >I am a link but looks like a button</a
```
rebecca-shoptaw, line 35: Tiny typo here -- should be `I am a link but look like a button`

packages/ia-styles/index.ts
```diff
@@ -0,0 +1,3 @@
+export { iaButtonStyles } from './src/ia-button';
+export { iaSronlyStyles } from './src/ia-sronly';
```
rebecca-shoptaw, line 2: I love how extensible this is!

rebecca-shoptaw (approved):
One note re: typo in the demo, but otherwise LGTM and will be a great help!

PR #945, opened 2025-03-20
WEBDEV-7393: Convert Wayback Search to TS and updates deps

jbuckner:


packages/ia-wayback-search/src/ia-wayback-search.ts
```diff
@@ -0,0 +1,201 @@
+import { css, html, LitElement } from "lit";
```
bfalling, line 1: QUESTION: Do we have a convention re: single quotes vs. double quotes. I had thought we always went with single quotes unless there was a reason to take advantage of the extra incremental richness of double quotes. (Kind of like defaulting to `const` and using `let` where needed.) But maybe I'm wrong.

jbuckner: Ah yeah.. not sure why the linter didn't catch that, but I added a rule to make sure it does.

iisa (approved):

PR #963, opened 2025-08-28
WBDEV-7675 Improve wayback search accessibility

rebecca-shoptaw:
Addresses two issues with the Wayback search flagged by Lighthouse: color contrast and missing aria label.


bfalling (approved):
LGTM

PR #993, opened 2026-03-12
Offshoot Topnav - Arrow keys not working correctly in topnav search f…

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-8237


jbuckner (approved):
One approach question, but looks good otherwise.

PR #997, opened 2026-04-15
WEBDEV-8238 Remove /account from signup links

rebecca-shoptaw:
Switches all signup links to `/signup` from `/account/signup` to prepare them to point to the new version of the page without redirects once it is released.

### DIFF
--- a/packages/ia-topnav/src/data/menus.ts
+++ b/packages/ia-topnav/src/data/menus.ts
@@ -638,7 +638,7 @@ export function buildTopNavMenus(
     ],
     signedOut: [
       {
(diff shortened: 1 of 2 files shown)


packages/ia-topnav/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/ia-topnav",
-  "version": "1.4.0",
+  "version": "1.4.1-webdev-8238.0",
```
rebecca-shoptaw: Will be reverted before merge

jbuckner (approved):

PR #1000, opened 2026-04-28
[ia-topnav] WEBDEV-8259 Switch to slotted search bar

latonv:
The existing topnav search bar is being replaced by a newer version with a dropdown menu more closely matching our search page tabs. This new search bar component is being slotted in from the parent app-root.

This PR replaces much of the existing nav-search setup with the new, much simpler slot approach, leaving the actual implementation and search functionality up to the slotted component.


packages/ia-topnav/src/ia-topnav.ts
```diff
@@ -86,37 +80,32 @@ export class IATopNav extends LitElement {
...
-          this.openMenu,
-        );
-
-        if (this.previousKeydownListener) {
-          this.removeEventListener('keydown', this.previousKeydownListener);
-        }
-        this.addEventListener('keydown', this.keyboardNavigation.handleKeyDown);
-        this.previousKeydownListener = this.keyboardNavigation.handleKeyDown;
-      }
+  firstUpdated() {
```
jbuckner: Should we do this in `connectedCallback` since that's when it's been attached to the DOM?

jbuckner (approved):
One question, but looks good! Cool how many files this removed!

packages/ia-topnav/src/ia-topnav.ts
latonv: Ah yeah, that would make sense. I'll swap that in, thanks!

PR #1004, opened 2026-05-05
[ia-topnav] WEBDEV-8396 Fix account menu positioning

latonv:
Currently the account menus (both logged-in and logged-out versions) are showing their dropdown menus at the wrong horizontal positions. This appears to be because the offsets are hardcoded based on the old size/positioning of the search bar, which recently changed. This PR updates the menu positioning to use a less brittle, dynamic approach that should always align it correctly with the toggle button.


packages/ia-topnav/src/styles/signed-out-dropdown.ts
latonv, line 1: (The only purpose of these deleted files was to hard-code menu offsets for various cases which are no longer relevant).

packages/ia-topnav/src/dropdown-menu.ts
```diff
@@ -14,16 +14,15 @@ import KeyboardNavigation from './lib/keyboard-navigation';
...
   @property({ type: Array }) menuItems: IATopNavLink[] | IATopNavLink[][] = [];
   @property({ type: Boolean }) animated = false;
   @property({ type: Boolean }) open = false;
 
   private previousKeydownListener?: // eslint-disable-next-line @typescript-eslint/no-explicit-any
   (this: HTMLElement, ev: KeyboardEvent) => any;
 
-  static get styles(): CSSResult[] {
-    return [dropdownMenuCSS];
+  static get styles() {
```
iisa: did we want return type value here or is it inherited?

iisa (approved):

packages/ia-topnav/src/dropdown-menu.ts
latonv: Ah, good call. Updated to `CSSResultGroup` which should be more future-proofed whether we do/don't need to return an array here in the future.

PR #1007, opened 2026-05-12
Rewrite & fix ia-zendesk-help-widget in iaux repo with TS/WC

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-8356


packages/ia-zendesk-help-widget/demo/app-root.ts
```diff
@@ -0,0 +1,21 @@
+import { html, css, LitElement } from 'lit';
+import { customElement } from 'lit/decorators.js';
+import '../src/ia-zendesk-help-widget';
+
+@customElement('app-root')
+export class AppRoot extends LitElement {
+  render() {
+    return html`
+      <ia-zendesk-help-widget
+        .widgetKey=${'6fe87bd8-d4e3-4b42-8632-be6eb933d54d'}
```
jbuckner, line 10: Is this a real key? ( might not use it in the demo if so)

packages/ia-zendesk-help-widget/demo/index.html
```diff
@@ -0,0 +1,21 @@
+<!doctype html>
+<html lang="en-GB">
+<head>
+  <meta name="viewport" content="width=device-width,initial-scale=1">
+  <meta charset="utf-8">
+  <style>
+    html {
+      font-size: 10px; /* This is to match petabox's base font size */
```
jbuckner: Does this make a difference for this repo/demo?

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.ts
```diff
@@ -0,0 +1,192 @@
...
+ *     --button-right: 20px;
+ *   "
+ * ></ia-zendesk-help-widget>
+ * ```
+ */
+@customElement('ia-zendesk-help-widget')
+export class IAZendeskHelpWidget extends LitElement {
+  /** Zendesk account key from the `ze-snippet` URL. */
+  @property({ type: String })
+  widgetKey = '6fe87bd8-d4e3-4b42-8632-be6eb933d54d';
```
jbuckner: We probably shouldn't provide a default key in the element. Just let the consumer provide one.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  /** Viewport width (px) below which the label is automatically hidden. */
+  @property({ type: Number })
+  breakpoint = 767;
```
jbuckner: Is this to maintain customizability of when the label appears? I wonder if there's a pure CSS solution to this so we're not having to inject TS into the CSS 🤔

packages/ia-zendesk-help-widget/demo/app-root.ts
nsharma123: This is test key from my personal account.

packages/ia-zendesk-help-widget/demo/index.html
nsharma123: Removed.

packages/ia-zendesk-help-widget/src/ia-zendesk-help-widget.ts
nsharma123: The other solution was to use a media query to suppress the label below 767px. Since we were already using a media query at 767px earlier, I updated it to keep the behavior consistent with the previous implementation.

packages/ia-zendesk-help-widget/.DS_Store
jbuckner, line 1: We probably shouldn't include `.DS_Store` files in the repo

packages/ia-zendesk-help-widget/src/zendesk-service.ts
```diff
@@ -0,0 +1,55 @@
...
+    zE?: {
+      (
+        target: 'messenger:on',
+        event: 'open' | 'close',
+        callback: () => void,
+      ): void;
+      (target: 'messenger', action: 'open' | 'close'): void;
+    };
+  }
+}
```
jbuckner: Can we pull this out to its own file?

jbuckner (approved):
I don't think I can fully review this due to the size of the MR, but it looks like its working in the review app

PR #1012, opened 2026-07-13
WEBDEV-8723 Switch topnav to new logout URL

rebecca-shoptaw:
Switches `/account/logout` to `/logout` in the topnav menu


pezvi (approved):
LGTM

PR #1014, opened 2026-07-18
WEBDEV-8763: Export radio-player theater building blocks from the package root

jbuckner:
`index.js` already re-exports these at runtime; `index.d.ts` didn't. This syncs the type declarations so consumers can import `AudioSource`, `TranscriptConfig`, `TranscriptEntryConfig`, `SearchHandler`, `LocalSearchBackend`, `FullTextSearchBackend`, `TranscriptIndex`, the full-text-search response types, and `SearchHandlerInterface` from the package root instead of reaching into the transitive `audio-element` / `transcript-view` packages or `radio-player/lib/src` deep paths.

Type-only, no runtime change. Bumps 1.0.0 to 1.0.1.

Unblocks the offshoot theater-player decoupling (WEBDEV-8762). Validated by pointing offshoot at a patched copy carrying these re-exports: it typechecks importing everything from the root.

Didn't build locally (repo isn't installed in my checkout); CI covers the build.

### DIFF
--- a/packages/radio-player/index.d.ts
+++ b/packages/radio-player/index.d.ts
@@ -1,15 +1,40 @@
+import {
+  TranscriptConfig,
+  TranscriptEntryConfig
+} from '@internetarchive/transcript-view';
+import { AudioSource } from '@internetarchive/audio-element';
+
 import RadioPlayer from './lib/src/radio-player.d';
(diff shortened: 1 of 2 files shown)


bfalling (approved):
LGTM

PR #1018, opened 2026-08-18
WEBDEV-8889: Fix transcript-view search results never scrolling into view

jbuckner:
`selectedSearchResult` looks the entry up with an attribute selector, but the template bound `data-search-result-index` as a Lit property, so the attribute was never written and the getter always returned null. The transcript stayed at the top on load and the `<` `>` switcher arrows did nothing. Dropping the dot fixes it.

Came out of debugging the offshoot radio player (WEBDEV-8886). petabox is unaffected because it's still on transcript-view 0.0.3, where the same line is a plain attribute binding.

Bumps 1.0.0 to 1.0.1.

## Verification

New test fails on master (`expected null to equal '0'`) and passes with the fix. Full package suite: 25 passing.

Also confirmed end to end by patching the built package into offshoot and loading a real radio item: transcript `scrollTop` went 0 to 2177, matching petabox exactly, and the `>` arrow advanced it to 3738.

Note `npm test` (with `--coverage`) can't run in this package right now: babel chokes on optional chaining in `@open-wc/testing-helpers`. Pre-existing on master, unrelated to this change. Ran `karma start` without coverage instead.

https://claude.ai/code/session_01HGzyVutjSEETR8wtd4YExL

### DIFF
--- a/packages/transcript-view/package.json
+++ b/packages/transcript-view/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/transcript-view",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "A Transcript View LitElement that displays closed captioning and search results.",
   "license": "AGPL-3.0-only",
   "main": "index.js",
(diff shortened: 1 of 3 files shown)


latonv (approved):
LGTM, no blockers

PR #1019, opened 2026-08-24
WEBDEV-8937: Require transcript-view ^1.0.1

jbuckner:
Bumps the range so the WEBDEV-8889 scroll fix is required, not just available. A fresh resolve already gets 1.0.1, but a consumer with a lockfile pinned to 1.0.0 keeps the broken build until someone refreshes it by hand (that's WEBDEV-8886).

Also bumps radio-player to 1.0.2 for the release.

https://claude.ai/code/session_016sdEUHvndFvK8qvG8ZLm6k

### DIFF
--- a/packages/radio-player/package.json
+++ b/packages/radio-player/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@internetarchive/radio-player",
-  "version": "1.0.1",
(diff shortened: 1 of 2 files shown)


pezvi (approved):
lgtm

